PR #154, opened 2023-02-28
Add docker-compose to spin up cardanon-node

### LINKED ISSUE #90: Use docker-compose for starting 3 Hydra nodes and one common Cardano node
Plan is to mainly reuse code from hydra-demo.

We also need to make sure that keys our CLI use are matching one used by Hydra nodes so this task is related to #87 .

nazrhom:
Closes #90


GeorgeFlerovsky:
@uhbif19 Should we close this for now?

nazrhom:
@GeorgeFlerovsky why close it? I though we had agreed to go with compose

nazrhom:
I added the hydra-nodes and a script that starts everything and seeds the actors running the nodes with some ada and some fuel

GeorgeFlerovsky:
Ah, OK. Nevermind. I got confused 😅

I'll go ahead and review it then 👍

stop-demo.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh -e
+
+docker-compose down
+docker-compose --profile hydra-node down
```
uhbif19, line 4: Why we need profiles here?

Did not know about them BTW.

docker-compose.yaml
```diff
@@ -2,19 +2,138 @@ version: "3.9"
 
 services:
   cardano-node:
-    image: inputoutput/cardano-node:1.35.4
+    image: inputoutput/cardano-node:latest
```
uhbif19, line 5: I am not sure if use latest is good, cuz `cardano-node` might break, while I guess we want need to change version often.

uhbif19:
```suggestion
    image: inputoutput/cardano-node:1.35.4
```

data/credentials/generate-actor.sh
```diff
@@ -0,0 +1,3 @@
+cardano-cli address key-gen \
```
uhbif19, line 1: I do not think that this is `data`.

data/hydra-keys/oscar.sk
```diff
@@ -0,0 +1,5 @@
+{
```
uhbif19, line 1: I think that situation with those actors should be covered in docs somethere.

uhbif19 (changes requested):

stop-demo.sh
nazrhom: We use the profile to start the hydra-nodes after the cardano-node (even though they are all in the same dockerfile)

seed-devnet.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
nazrhom: Yes, this step needs to happen to give the actors running the hydra nodes enough funds to manage the hydra lifecycle

docker-compose.yaml
```diff
@@ -2,19 +2,138 @@ version: "3.9"
...
-      ]
\ No newline at end of file
+      , "--shelley-kes-key", "/devnet/kes.skey"
+      , "--shelley-vrf-key", "/devnet/vrf.skey"
+      , "--shelley-operational-certificate", "/devnet/opcert.cert"
+      , "--byron-delegation-certificate", "/devnet/byron-delegation.cert"
+      , "--byron-signing-key", "/devnet/byron-delegate.key"
+      ]
+
+  hydra-node-1:
```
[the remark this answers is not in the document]
nazrhom: TBH I don't think it is worth having them for such a contained file, they make the document harder to read and reason about (also I never see them used).
Also, most keys have slight differences, so we would really only use them for the volumes and profiles

scripts/reset-devnet.sh
```diff
@@ -4,15 +4,16 @@
 # "up-to-date" genesis files. If the directory exists, it is wiped out.
 set -e
 
-TARGETDIR="devnet"
+ROOTDIR="."
```
uhbif19, line 7: Nice.

scripts/spin-up-new-devnet.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/sh -e
+parent_path=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
```
uhbif19, line 2: Do not understand why this is required, but okay.

stop-demo.sh
uhbif19: Does not `depends_on` does this?

While I am a little bit out of sync with latest docker changes, so I may miss something.

uhbif19 (approved):

PR #171, opened 2023-03-07
Remove utxo parameter from announce-auction. The CLI will now look in…

### LINKED ISSUE #113: Select nonce UTxO automatiaclly
> @GeorgeFlerovsky #44 is about user errors too. Then you provide wrong nonce UTxO or taking some step then no proper UTxOs for that step is present.

Actually, for the demo, I would like to remove the need for the CLI user to specify the nonce utxo.
(Maybe this should be a separate GH issue)

It would be cleaner to:

- Search the user's utxos for the "Mona Lisa" token.
- Set the `utxoNonce` to be the TxOutRef of the utxo that contains the token.
- If the user doesn't have the token, then give a user error 'Selected user doesn't have the "Mona Lisa" token.This demo is configured to use this token as the auction lot.'

The more general mechanism to select arbitrary auction lot tokens will be needed in production, but it's kind of irrelevant to our development for now. On the other, requiring users to copy-paste utxo refs in CLI is really ugly UX — reminds me of the early Alonzo days.

When we do need this general functionality in the future, we'll probably be replacing the CLI interface with a wallet-based web interface, which has a better UX for selecting tokens/utxos.

_Originally posted by @GeorgeFlerovsky in https://github.com/mlabs-haskell/hydra-auction/issues/108#issuecomment-1433578595_

nazrhom:
… the user UTxOs for a token matching the Mona Lisa asset.

Closes #113


app/CLI/Actions.hs
```diff
@@ -126,18 +126,22 @@ handleCliAction userAction = do
...
-      Just config <- liftIO $ readAuctionTermsConfig auctionName
-      terms <- liftIO $ configToAuctionTerms config dynamic
-      liftIO . putStrLn $
-        show actor
-          <> " announces auction called "
-          <> show auctionName
-          <> "."
-      announceAuction terms
+    AuctionAnounce auctionName -> do
+      mUtxo <- findTestNFT <$> actorTipUtxo
```
uhbif19, line 130: Guess our naming was wrong, cuz it is TxIn, not UTxO.

uhbif19 (approved):

uhbif19:
@nazrhom Sorry, did not know about that feature. Thanks for explanation.

PR #183, opened 2023-03-10
Add HLS starting script

uhbif19:


MangoIV:
This will go wrong. Do you not have HLS available in the devShell? I.e. what does haskell-language-server-wrapper —-version say?

MangoIV:
yeah, this works perfectly fine for me: 
image

scripts/run-hls.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+nix develop --command $(find /nix/store -maxdepth 1 -type d -iname '*haskell-language-server-exe*' | head -n 1| xargs -I % sh -c 'find  % -type f -name "haskell-language-server"') --lsp
```
MangoIV, line 2: this will ultimately fail; the nix store can contain as many language-server versions as we want; there is no reason for this command to find the one we need. haskell-language-server is very fragile when it comes to compatibility, even for the same ghc version there can be ABI mismatches; when in the `devShell`, simply calling `haskell-language-server-wrapper --lsp` should suffice.

MangoIV (changes requested):

scripts/run-hls.sh
uhbif19: This does not works for me and fails with error.

And: 

```
> whereis haskell-language-server                                                        1 ✘  9s   impure  
haskell-language-server: /nix/store/q4jv52s81yln9r2x40qq31fs7x8lwp04-haskell-language-server-exe-haskell-language-server-1.8.0.0/bin/haskell-language-server
```

I am starting shell with `nix develop` do you do it some other way?

uhbif19: Maybe I could just do `whereis` in `nix develop --command`?

MangoIV: `nix develop -c bash -c "haskell-language-server-wrapper --lsp"` should do what you want.

MangoIV: `haskell-language-server-wrapper` manages available `haskell-language-server` binaries, it will pick the correct one based on the environment you're in. It will try to pick one for the current project and will fallback to whatever is left over (in your case the one globally installed by ghcup); that's why you want to run `nix develop` first. Consider using `direnv`, this will significantly reduce the overhead you have with `nix` 

(nix profile install direnv) `echo "use flake -Lv --fallback" | save .envrc; direnv allow`

uhbif19: `nix develop --command $(which haskell-language-server) --lsp` works in CLI, but does not work in VS Code extension. Not sure why.

MangoIV: > As I said `haskell-language-server-wrapper` is not present in my nix develop.

Is it present outside of it? I think it's part of the global `hls` installation

MangoIV: > nix develop --command $(which haskell-language-server) --lsp works in CLI, but does not work in VS Code extension. Not sure why.

perhaps because the `executable` is not called that? Perhaps `"haskell-language-server-8.10.7"`? 

This is what my invocation of `hls-wrapper` spits out: 

```
Found "/home/mangoiv/Work/mlabs/hydra-auction/hie.yaml" for "/home/mangoiv/Work/mlabs/hydra-auction/a"
Run entered for haskell-language-server-wrapper(haskell-language-server-wrapper) Version 1.9.1.0 x86_64 ghc-9.4.4
Current directory: /home/mangoiv/Work/mlabs/hydra-auction
Operating system: linux
Arguments: []
Cradle directory: /home/mangoiv/Work/mlabs/hydra-auction
Cradle type: Cabal

Tool versions found on the $PATH
cabal:          3.8.1.0
stack:          2.9.1
ghc:            8.10.7


Consulting the cradle to get project GHC version...
Project GHC version: 8.10.7
haskell-language-server exe candidates: ["haskell-language-server-8.10.7","haskell-language-server"]
Launching haskell-language-server exe at:/nix/store/q4jv52s81yln9r2x40qq31fs7x8lwp04-haskell-language-server-exe-haskell-language-server-1.8.0.0/bin/haskell-language-server
2023-03-10T15:52:15.188195Z | Info | No log file specified; using stderr.
2023-03-10T15:52:15.188872Z | Info | haskell-language-server version: 1.8.0.0 (GHC: 8.10.7) (PATH: /nix/store/q4jv52s81yln9r2x40qq31fs7x8lwp04-haskell-language-server-exe-haskell-language-server-1.8.0.0/bin/haskell-language-server)
2023-03-10T15:52:15.189259Z | Info | Directory: /home/mangoiv/Work/mlabs/hydra-auction
2023-03-10T15:52:15.189423Z | Info | Logging heap statistics every 60.00s
 ghcide setup tester in /home/mangoiv/Work/mlabs/hydra-auction.
Report bugs at https://github.com/haskell/haskell-language-server/issues
```

MangoIV (approved):
I still don't know why `hls-wrapper` doesn't work here, but I guess this is fine if it works; if you wish you may perhaps add something like "If you're having trouble using hls with this project" :)

uhbif19:
> if you wish you may perhaps add something like 

Added.

README.md
```diff
@@ -91,6 +91,10 @@ To run app with GHC warnings present you can use:
 cabal run --ghc-option='-Wwarn'
 ```
 
+If youa are having trouble using hls with this project, you can use following
```
MangoIV, line 94:
```suggestion
If you are having trouble using hls with this project, you can use following
```

PR #185, opened 2023-03-10
[chore] refactor common in submodules

### LINKED ISSUE #100: Refactor `HydraAuction.Tx.Common`
* It is too big and have unrelated functions
* It mixes utility functions and functions actually related to our project

MangoIV:
- Runner: everything related to the Runner type
- Types: Common Types (at the moment just one)
- Utils: Utility functions that are not specific to the Runner Type
- Common: reexports everything from Runner, Types and Utils


uhbif19 (changes requested):
This PR goes against the solution proposed in issue and I do not see reasoning why so.

Also I do not understand why integrate `Utils` in `Commons` while taking out domain-independent Utils from Commons was the whole point of issue.

Also I think that placing everything working with Runner in Runner module is bad, because there is a lot of different stuff working on top Runner.

And last change highlights issues with Runner which was present before. Functions we do should be domain independent. While `actor` in `ExecutionContext` seems not to be applicable to real-world projects. But I guess creating two versions of this monad may be overkill and we could just continue with current state.

Also `Runner` is bad name, cuz not explaining specifics of runner.

MangoIV:
Please excuse, I guess I didn't quite understand what the solution proposed was, I think I will have to have a look again. I thought the main idea was separation of domain specific logic vs general helpers. Sorry for the misconception.

uhbif19:
> the main idea was separation of domain specific logic vs general helpers

Yes, but I am do not see it in this PR, may be I am understanding something wrong.

MangoIV:
Most probably my misunderstanding, will fix. :)

MangoIV:
> And last change highlights issues with Runner which was present before. Functions we do should be domain independent. While actor in ExecutionContext seems not to be applicable to real-world projects. But I guess creating two versions of this monad may be overkill and we could just continue with current state.

do you want something like tagless final?

MangoIV:
yeah, I think this is similar to what I was thinking about, but more fine-grained; so you wouldn't have `MonadRunner` but a subset of the constraints of `MonadRunner` which don't include `actor`

uhbif19:
Do not know why github shows a lot of unrelated commits while seems like they all was merged to staging. Maybe this is due to some rebasing?

PR #187, opened 2023-03-10
[chore] update haskellNix, change flake url to use https instead of ssh

### LINKED ISSUE #186: [chore] clean up nix flake and update tooling
- [ ] clean up flake 
- [ ] update tooling 

> **Note**
> This is low priority but should offer a quality of live improvement

MangoIV:
> **Warning**
> This is WIP, I just started by updating the `haskell.nix` flake input, that way, it can build 
> in our CI

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -30,7 +30,7 @@
     empty.url = "github:mlabs-haskell/empty-flake";
     iohk-nix.follows = "hydra/iohk-nix";
     CHaP.follows = "hydra/CHaP";
-    nixpkgs.follows = "hydra/nixpkgs";
+    nixpkgs.follows = "haskellNix/nixpkgs";
     flake-utils.follows = "hydra/flake-utils";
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -17,7 +17,8 @@
     # when you upgrade `hydra` input remember to also upgrade revs under
     # `source-repository-package`s in `cabal.project`
     hydra = {
-      url = "ssh://git@github.com/input-output-hk/hydra?ref=5ed00dfcd367d0390a774216035e1ea30dde5166";
```
nazrhom, line 20: I seem to recall we had to use ssh here as it was the only way to get it to clone with submodules, but maybe I am wrong

MangoIV: That sounds so absurd that I would expect it to be true in nix land. At least it didn't reject it as invalid, so perhaps it worked? I'll have to find out.

PR #192, opened 2023-03-14
Gf/watch auction

### LINKED ISSUE #131: Notify user on auction state changes in CLI
Monitor time in separate process and then time pass certain stage change - notify user.

nazrhom:


app/CLI/Watch.hs
```diff
@@ -28,19 +34,25 @@ watchAuction auctionName = do
...
+          <> show currentStage
+          <> case mSecsLeft of
+            Nothing -> mempty
+            Just s -> "\n" <> show s <> " seconds left until next stage"
 
   threadDelay 0.2
   watchAuction auctionName
+
+secondsLeftInInterval :: Integer -> Interval POSIXTime -> Maybe Integer
+secondsLeftInInterval now (Interval _ (UpperBound (Finite (POSIXTime t)) inclusive)) = Just $ (t - now - if inclusive then 0 else 1) `div` 1000
```
uhbif19, line 57: Why you substract 1 but then `div` for 1000? Or div will round that 1 automatically? Probably you right.

uhbif19 (changes requested):
I think this should have tests for seconds calculation, cuz arithmetics could be tricky in such cases.

uhbif19:
I thought that plan was to add `seconds left till correct stage` for `wrong stage` error too, but I guess, that in case we have `watch` command it is already clear.

app/CLI/Watch.hs
nazrhom: div already does the rounding. it is what is usually known as integer division

src/HydraAuction/OnChain/Common.hs
```diff
@@ -60,6 +62,11 @@ checkVoucherExpiredOrLater terms info =
   traceIfFalse "Wrong interval for transaction" $
     contains (from (voucherExpiry terms)) (txInfoValidRange info)
 
+-- This function is not used on-chain
+secondsLeftInInterval :: Integer -> Interval POSIXTime -> Maybe Integer
```
MangoIV, line 66: I think either this should get haddock or the first parameter's type should be changed to be a type that represents time, especially as this makes assumptions about the type of argument that we get passed. Ideally the type of the first argument should be changed as well.

app/CLI/Watch.hs
```diff
@@ -0,0 +1,50 @@
...
+      let showTime = formatTime defaultTimeLocale "%Y-%m-%d %H:%M:%S"
+          mSecsLeft = secondsLeftInInterval currentTimeMs (stageToInterval terms currentStage)
+      putStrLn $ showTime currentTime
+      putStrLn $ "Auction: " <> show auctionName
+      putStrLn $ "Seller: " <> show sellerActor
+      putStrLn $
+        "Current stage: "
+          <> show currentStage
+          <> case mSecsLeft of
+            Nothing -> mempty
```
MangoIV, line 46: this can perhaps be an error message, telling that the `Interval` was not constructed properly by the `stageToInterval` function

app/Main.hs
```diff
@@ -43,15 +38,23 @@ import CardanoNode (
...
 
 main :: IO ()
 main = do
-  MkCliInput {cliVerbosity, cliActor, cliNodeSocket, cliNetworkId} <- getCliInput
+  ci <- getCliInput
+  handleCliInput ci
 
+handleCliInput :: CliInput -> IO ()
+handleCliInput (Watch auctionName) = watchAuction auctionName
+handleCliInput (InteractivePrompt MkPromptOptions {..}) = do
```
MangoIV, line 57: Just for visibility I want to repeat what I said on the other review, namely that I think for records with many fields this is not very nice as it makes looking up the record fields very hard.

test/Unit/Common.hs
```diff
@@ -49,10 +53,10 @@ testCurrentAuctionStage = do
...
   let config =
         AuctionTermsConfig
-          { configDiffBiddingStart = 1 * 60 - 1
-          , configDiffBiddingEnd = 2 * 60 - 1
-          , configDiffVoucherExpiry = 3 * 60 - 1
-          , configDiffCleanup = 4 * 60 - 1
+          { configDiffBiddingStart = 1 * 60 - 2
+          , configDiffBiddingEnd = 2 * 60 - 2
+          , configDiffVoucherExpiry = 3 * 60 - 2
+          , configDiffCleanup = 4 * 60 - 2
```
MangoIV, line 59: Perhaps it can be documented why we choose to have these slightly weird times.

src/HydraAuction/OnChain/Common.hs
MangoIV, line 66:
```suggestion
-- | Given a time in Milliseconds, and an 'Interval' this function computes 
--   the truncated difference in seconds to the 'UpperBound' of the passed 'Interval'.  
--   The 'Interval' should always have a finite 'Upperbound'. 
--   Note that this function is not and should not be used on-chain
secondsLeftInInterval :: Integer -> Interval POSIXTime -> Maybe Integer
```

MangoIV: perhaps like this?

app/CLI/Watch.hs
nazrhom: That is not the case though, the idea is that once you are in the `CleanupStage` there is no upperbound to the validity interval, and no amount of seconds left for the next stage.

test/Unit/Common.hs
nazrhom: Which part (or both) do you think we should explain better?

src/HydraAuction/OnChain/Common.hs
nazrhom, line 66:
```suggestion
-- | Given a time in Milliseconds, and an 'Interval' this function computes 
--   the truncated difference in seconds to the 'UpperBound' of the passed 'Interval'.  
--   If the Interval does not have a finite `UpperBound` the function will return Nothing.
--  This function also assumes that the given time in Milliseconds is smaller than the `UpperBound` of the interval (if that exists, and is finite)
--   Note that this function is not and should not be used on-chain
secondsLeftInInterval :: Integer -> Interval POSIXTime -> Maybe Integer
```

src/HydraAuction/OnChain/Common.hs
```diff
@@ -60,6 +62,11 @@ checkVoucherExpiredOrLater terms info =
   traceIfFalse "Wrong interval for transaction" $
     contains (from (voucherExpiry terms)) (txInfoValidRange info)
 
+-- This function is not used on-chain
```
nazrhom, line 65:
```suggestion
-- | Given a time in Milliseconds, and an 'Interval' this function computes 
--   the truncated difference in seconds to the 'UpperBound' of the passed 'Interval'.  
--   If the Interval does not have a finite `UpperBound` the function will return Nothing.
--  This function also assumes that the given time in Milliseconds is smaller than the `UpperBound` of the interval (if that exists, and is finite)
--   Note that this function is not and should not be used on-chain
```

nazrhom: @MangoIV what about this?

app/Main.hs
nazrhom: While I don't personally agree with the hate `RecordWildcards` gets, it is worth mentioning that we explicitly mention not to use it in the styleguide https://github.com/mlabs-haskell/hydra-auction/blob/staging/doc/contributing.md#other-guidelines yet have it enabled in several files through pragmas 😅

nazrhom: Part of the issue, is we say: don't use it, except if you really need to. which is too vague to be enforceable, maybe we can come up with a concrete number of fields for which we allow it (or just ban it completely!)

src/HydraAuction/OnChain/Common.hs
MangoIV: looks good to me :)

app/CLI/Watch.hs
MangoIV: and that's the only codepath in all situations that can lead to `Nothing` there?

src/HydraAuction/OnChain/Common.hs
```diff
@@ -60,6 +62,15 @@ checkVoucherExpiredOrLater terms info =
   traceIfFalse "Wrong interval for transaction" $
     contains (from (voucherExpiry terms)) (txInfoValidRange info)
 
+-- | Given a time in Milliseconds, and an 'Interval' this function computes
```
uhbif19: > Given a time in Milliseconds

Is not this just `POSIXTime`?

test/Unit/Common.hs
```diff
@@ -72,5 +76,16 @@ testCurrentAuctionStage = do
   assertStageAtTime terms (4 `minutes` later) CleanupStage
   where
     assertStageAtTime terms timeDiff expectedStage = do
-      _ <- travelTo timeDiff $ currentAuctionStage terms
-      liftIO $ expectedStage @=? expectedStage
+      stage <- travelTo timeDiff $ currentAuctionStage terms
+      liftIO $ stage @?= expectedStage
```
uhbif19, line 83: Oh, my bad.

test/Unit/Common.hs
```diff
@@ -72,5 +76,16 @@ testCurrentAuctionStage = do
...
+
+testSecondsLeftInterval :: Assertion
+testSecondsLeftInterval = do
+  let now = 1000
+      interval1 = to 5000
+      interval2 = interval 5000 15000
```
uhbif19, line 89: That does not cover case of `now` being after UpperBound. I guess in that case your `-1` trick should probably break. So I think function should just return Nothing, cuz it not supposed to work with this case at all.

uhbif19 (changes requested):

test/Unit/Common.hs
```diff
@@ -51,12 +51,15 @@ testCurrentAuctionStage = do
...
   let config =
         AuctionTermsConfig
-          { configDiffBiddingStart = 1 * 60 - 2
-          , configDiffBiddingEnd = 2 * 60 - 2
-          , configDiffVoucherExpiry = 3 * 60 - 2
-          , configDiffCleanup = 4 * 60 - 2
+          { configDiffBiddingStart = 30
+          , configDiffBiddingEnd = 90
+          , configDiffVoucherExpiry = 150
+          , configDiffCleanup = 210
```
MangoIV, line 62: this is nicer

MangoIV (approved):
Looks good to me :)

src/HydraAuction/OnChain/Common.hs
nazrhom: Yeah, actually lets just make the argument POSIXTime

test/Unit/Common.hs
uhbif19: I do not think that is clearer than arithmetic expression, cuz logic of number is missed, but okay.

uhbif19 (approved):

app/Main.hs
uhbif19: > Part of the issue, is we say: don't use it, except if you really need to. which is too vague to be enforceable, maybe we can come up with a concrete number of fields for which we allow it (or just ban it completely!)

Is >= 5 fields criteria okay?

PR #198, opened 2023-03-15
Change Add to project bot to one what uses new Projects API

uhbif19:


uhbif19:
`Act` check fails with error. Not sure if it is error in action or `act`.

nazrhom:
Hey @uhbif19 I noticed these changes are targeting master, is that intended?

uhbif19:
@nazrhom I think this workflows only work in master, no? Also previous version is already in master. And this is our change not affecting uses, so I see no reason not to put it in master.

MangoIV (approved):

uhbif19:
@MangoIV This cannot break CI check, AFAIK, as Github treating these workflows as separate things.

Issue #140, opened 2023-02-24
Do delegate server stub


uhbif19:
Probably we need different cabal executable for this.

I guess they will be also started in `docker-compose` as `hydra-nodes` #90 

Tasks:

- [x] Add pure logic FSM-like implementation getting pure `DelegateEvents` as input.
- [ ] Add new executable: it starts a multiple threads which push events into shared queue

Issue #168, opened 2023-03-06
Decide on measures to authenticate Delegate server clients


uhbif19:
We need to authenticate sellers to prohibit a bad actor to place wrong standing bid on L2.

We need to authenticate bidders to prohibit somebody to place bids on their behalf.
While this will probably be covered by signature required onchain.

Overall the scope of this task is to make a decision about implementation of this.

uhbif19:
> We need to authenticate sellers to prohibit a bad actor to place wrong standing bid on L2.

Seems like there is no need for this, cuz standing bid script checks for correctness of moving and any auction can be placed at any hydra head.

I thought there was a reason for restriction but cannot remember it now.

Issue #158, opened 2023-03-03
Implement Hydra node requests on delegate server


uhbif19:
Most of things should be already covered by #124 and #125 

Domain logic should be covered by other tasks.

Scope of this task is to actually react on `DelegateEvents` by triggering actions on L2. Concretelly:

- [ ] Introduce Hydra node and Cardano node connection in config or CLI option
- [x] Init/Abort/Close Head
- [x] Commit - both empty and StandingBid transactions case
- [x] Placing new bid
- [x] Probably testing this in E2E test?

No signing by bidder is required in this milestone.

Blocked by #140 
Blocked by #124 
Blocked by #125

PR #201, opened 2023-03-16
Delegate server and API stub - Take 2 (to fix strange conflicts)

uhbif19:
Closing #140

### DIFF
--- a/src/HydraAuction/Delegate.hs
+++ b/src/HydraAuction/Delegate.hs
@@ -2,6 +2,7 @@ module HydraAuction.Delegate (
   delegateStep,
   DelegateEvent (..),
   DelegateInput (..),
+  DelegateRunnerT (..),
 ) where
 
 -- Prelude imports
@@ -26,11 +27,7 @@ import HydraAuction.Delegate.Interface (
 import HydraAuction.Hydra.Interface (HydraEvent (..))
 import HydraAuction.Types (AuctionStage (..), AuctionTerms (..))
 
--- TODO: separate API and hydra
-
--- FIXME: how to track contenstation time
-
--- FIXME: add signature checking (where is issue on this)
+-- FIXME: add client authentication (where is issue on this)
 
 data DelegateEvent
   = Start
--- a/src/HydraAuction/Delegate.hs
+++ b/src/HydraAuction/Delegate.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE StrictData #-}
+
 module HydraAuction.Delegate (
   delegateStep,
   DelegateEvent (..),
--- a/src/HydraAuction/Delegate.hs
+++ b/src/HydraAuction/Delegate.hs
@@ -11,6 +11,7 @@ import Prelude
 
 import Control.Monad.Except (MonadError (throwError))
 import Control.Monad.State (MonadState, StateT (..), get, put)
+import Control.Monad.Trans (MonadTrans (lift))
 import Control.Monad.Trans.Except (ExceptT (..))
 
 -- HydraAuction imports
@@ -49,6 +50,9 @@ newtype DelegateRunnerT m x
     , MonadState DelegateState
     , MonadError DelegateError
     )
+
+instance MonadTrans DelegateRunnerT where
+  lift = MkDelegateRunner . lift . lift
 
 delegateStep ::
   forall m. Monad m => DelegateInput -> DelegateRunnerT m DelegateResponse
--- a/src/HydraAuction/Delegate.hs
+++ b/src/HydraAuction/Delegate.hs
@@ -10,20 +10,15 @@ module HydraAuction.Delegate (
 import Prelude
 
 -- Haskell imports
-import Control.Monad.Except (MonadError (throwError))
 import Control.Monad.IO.Class (MonadIO)
 import Control.Monad.State (MonadState, StateT (..), evalStateT, get, put)
 import Control.Monad.Trans (MonadTrans (lift))
-import Control.Monad.Trans.Except (runExceptT)
 
 -- HydraAuction imports
 
 import HydraAuction.Delegate.Interface (
-  DelegateError (..),
   DelegateResponse (..),
-  FrontendKind (..),
   FrontendRequest (..),
-  KnownFrontendRequest (..),
  )
 import HydraAuction.Hydra.Interface (HydraEvent (..))
 import HydraAuction.Types (AuctionStage (..), AuctionTerms (..))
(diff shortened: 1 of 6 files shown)


src/HydraAuction/Delegate.hs
```diff
@@ -0,0 +1,107 @@
...
+instance MonadTrans DelegateRunnerT where
+  lift = MkDelegateRunner . lift
+
+delegateStep ::
+  forall m.
+  Monad m =>
+  DelegateInput ->
+  DelegateRunnerT m (Either DelegateError DelegateResponse)
+delegateStep input = runExceptT $ case input of
+  DelegateEvent Start -> return Okay -- FIXME: initialize Hydra
```
nazrhom, line 69: I would propose to have the delegate server behave similarly to the CLI, in that you need to start the hydra-node before starting the server, and then pass in the http address on which it is listening on.

src/HydraAuction/Delegate/Interface.hs
```diff
@@ -0,0 +1,35 @@
...
+-- Prelude imports
+import Prelude
+
+-- Cardano imports
+import Cardano.Api.UTxO (UTxO)
+
+-- HydraAuction imports
+import HydraAuction.Types (AuctionTerms (..), Natural)
+
+data FrontendKind = Seller | Bidder
```
nazrhom, line 18: I would remove this distinction as well. It is out of scope to handle authentication between the bidders and the delegates, we can just trust what the CLI says for the demo.

src/HydraAuction/Delegate.hs
```diff
@@ -0,0 +1,107 @@
...
+
+data DelegateEvent
+  = Start
+  | AuctionStageStarted AuctionStage
+  | HydraEvent HydraEvent
+
+data DelegateInput
+  = DelegateEvent DelegateEvent
+  | FrontendRequest FrontendRequest
+data DelegateState = NoClient | HasClient FrontendKind AuctionTerms
```
nazrhom, line 41: Why don't we just bundle the `AuctionTerms` in each request the client makes? This way the delegate is not tied to a specific auction, but can handle more than one. This would also simplify the issue with threads and state.

src/HydraAuction/Delegate/Interface.hs
```diff
@@ -0,0 +1,35 @@
...
+data KnownFrontendRequest
```
nazrhom, line 19: We should identify the actor that is making the request as well

src/HydraAuction/Delegate.hs
uhbif19: Yes, that is idea. But this is delegated into #158 

The idea is to change `Monad m` into `MonadHydra m` then.

src/HydraAuction/Delegate/Interface.hs
uhbif19: Yes this is the datatype which defines who is client trusting CLI. Not trusting is postponed to another issue #168 

All logic code required for trusting-client-case is already in this PR.

src/HydraAuction/Delegate.hs
uhbif19: > This way the delegate is not tied to a specific auction.

It is tied in scope of our milestones. And probably will stay so until Hydra will support partial commiting/decommiting or something like that.

For case of multiple auctions we could just add auction id handle to requests and this still will work.

Giving AuctionTerms every time complicates further implementation of checking of client authenticity, adds possibility of changing AuctionTerms for hacking or just providing strange errors (cuz delegate wont find UTxO by changed AuctionTerms).

src/HydraAuction/Delegate/Interface.hs
uhbif19: Why? Seller only could be the one actor. Bidders provide their signatures on with newBid command.

Will think of that further.

src/HydraAuction/Delegate.hs
nazrhom: Why don't we remove the `Start` event then?

src/HydraAuction/Delegate/Interface.hs
nazrhom: What I mean is we could just ignore client authentication with the delegate server for the demo.

src/HydraAuction/Delegate.hs
nazrhom: I think we should favour simplicity over safety in this context, checking client authenticity is extra work that does not add much hydra-specific value.

nazrhom: Also, I don't think `NoClient/HasClient` is necessarily a good representation of the delegate state. Delegate servers are only there to interact with auctions, so what does a delegate server in the `NoClient` state really do? (other than wait for a client?).

src/HydraAuction/Delegate/Interface.hs
nazrhom: Actually you are right (I forgot bidders submitted signed txs), but then we should change the signature of NewBid from taking a Natural to taking the full tx then.

uhbif19: Yep, that is in Milestone 6. It is separated to issue #203 . Actually I wanted to discuss if you wanna take this issue next, cuz seems that M4.2 work will be splitted between me and @MangoIV

src/HydraAuction/Delegate.hs
uhbif19: Start event is required to Init Hydra head, which does not require to know Auction.

src/HydraAuction/Delegate/Interface.hs
uhbif19: We already do that, my code not requires to actually authenticate client, only store it role.

Anyway, we have a discussion with @GeorgeFlerovsky and found out that I was wrong, and it seems like there is no reasons to authenticate client at all or limit delegate to having one client. So I gonna change state to only store `AuctionTerms`.

src/HydraAuction/Delegate.hs
uhbif19: Yes, it waits for the client, and this is quite correct state, cuz delegate should exist before auction has started.

As for real work - delegate should do `Init` before having auction. And proposed future direction for `hydra-auction` is about having multiple auctions on single persisting Head, and thread only start working when Auction is given does not align with that.

src/HydraAuction/Delegate.hs
nazrhom: Sorry, I thought this was starting the hydra node instead, makes sense now!

nazrhom (approved):

PR #207, opened 2023-03-17
Add Cspell config

uhbif19:


nazrhom:
Hey @uhbif19 how do you run this? Should we add cspell to the nix tools?

uhbif19:
I am using Cspell as VS Code extension. Given what I am the main source of typos in this repo I guess this may be enough.

We can also add this to CI later, yes.

nazrhom:
We can add it to CI later, but I would add it to the nix shell right away. Maybe @MangoIV can help out with this?
Anyway, feel free to merge this PR in the meantime, no reason to block on this

nazrhom (approved):

PR #208, opened 2023-03-17
Check bidders are in the approvedBidders list when a new tx is submitted

### LINKED ISSUE #214: Check bidder is in `approvedBidders`

nazrhom:


README.md
```diff
@@ -37,19 +37,19 @@ Will build the CLI application and link it to `./result/bin/hydra-auction`.
...
 We recommend using the `spin-up-new-devnet.sh` script
 as it will take care of setting up the correct files for the cardano-node
 and seeding the actors running the hydra-nodes with enough funds to manage the head lifecycle.
 
 ## CLI usage
 
-Different terms for auction are stored in JSON and auctions named with string.
-The static parts of parameters like the timing of stages and minimal bid as well as
+Different terms for auction are stored in JSON and auctions are named with string.
+The static parts of parameters: like the timing of stages and minimal bid, as well as
```
uhbif19, line 50: A am not particularly good with English grammar but seems like : is wrong here.

src/HydraAuction/Fixture.hs
```diff
@@ -62,6 +68,14 @@ readConfigFile source = do
   filename <- Pkg.getDataFileName ("data" </> source)
   BS.readFile filename
 
+getActorVkHash :: Actor -> IO PubKeyHash
```
uhbif19, line 71: Nice.

src/HydraAuction/Fixture.hs
```diff
@@ -62,6 +68,14 @@ readConfigFile source = do
...
+getActorVkHash actor = do
+  (actorVk, _) <- keysFor actor
+  return $ toPlutusKeyHash $ verificationKeyHash actorVk
+
+getActorsPubKey :: [Actor] -> IO [PubKeyHash]
+getActorsPubKey actors = sequence $ getActorVkHash <$> actors
```
uhbif19, line 77: Does not look like something that should be in separate function to me.

src/HydraAuction/OnChain/StandingBid.hs
```diff
@@ -53,22 +56,23 @@ mkStandingBidValidator terms datum redeemer context =
...
         && case oldBid of
-          Bid oldBidTerms ->
+          Just oldBidTerms ->
             traceIfFalse "Bid increment is not greater than minimumBidIncrement" $
               bidAmount oldBidTerms + minimumBidIncrement terms <= bidAmount newBidTerms
-          NoBid ->
+          Nothing ->
             traceIfFalse "Bid is not greater than startingBid" $
               startingBid terms <= bidAmount newBidTerms
-    validNewBid _ NoBid = False
```
uhbif19, line 68: Why remove constructor into wildcard? Wilcard is less specific => more prone to miss logic error on function change.

src/HydraAuction/Tx/Escrow.hs
```diff
@@ -145,8 +140,8 @@ announceAuction terms = do
         , validityBound = (Nothing, Just $ biddingStart terms)
         }
 
-startBidding :: AuctionTerms -> Runner ()
-startBidding terms = do
+startBidding :: AuctionTerms -> ApprovedBidders -> Runner ()
```
uhbif19, line 121: Spec also prescribes recording `ApprovedBiddersHash` here AFAIU.

test/EndToEnd/Ledger.hs
```diff
@@ -168,11 +170,43 @@ sellerBidsTest = mkAssertion $ do
   announceAuction terms
 
   waitUntil $ biddingStart terms
-  startBidding terms
+  startBidding terms (ApprovedBidders [])
```
uhbif19: That is wrong list for `sellerBidsTest` I guess.

uhbif19: Also I think the `sellerBidsTest` tests now should check different thing - that seller cannot be added to the list of bidders.

That once this done he cannot be bidder is already checked by the `unauthorisedBidderTest` test.

src/HydraAuction/Tx/Common.hs
```diff
@@ -292,6 +299,32 @@ scriptUtxos script terms = do
...
+    [(_, out)] -> case txOutDatum out of
+      TxOutDatumInline scriptData ->
+        case fromData $ toPlutusData scriptData of
+          Just standingBidDatum -> standingBidDatum
+          Nothing ->
+            error "Impossible happened: Cannot decode standing bid datum"
+      _ -> error "Impossible happened: No inline data for standing bid"
+    _ -> error "Wrong number of standing bid UTxOs found"
+
+currentWinningBidder :: AuctionTerms -> Runner (Maybe PubKeyHash)
```
uhbif19: It is strange that this function was in `Escrow`, but still is not `StandingBid.hs` better place for it? Cuz it is tied to logic of StagingBid. Same with other two.

uhbif19 (changes requested):
Sorry for late review.

Good changes, but for closing issue PR is missing:

* `ApproverBiddersHash` from spec
* Checking bidders signatures from issue

src/HydraAuction/Tx/Escrow.hs
nazrhom: I'll add that, tbh I realised I wasn't 100% sure why we have it. The spec says it is to avoid with tampering, but when would that happen?
We already enforce that the standing bid does not change thelist of approved bidders in its validator

test/EndToEnd/Ledger.hs
nazrhom: That is a very good point! I'll update the test

src/HydraAuction/Tx/Common.hs
nazrhom: Unfortunately that gives rise to a circular dependency :(

src/HydraAuction/Tx/Escrow.hs
uhbif19: Hm, you are probably right. AFAI remember `ApprovedBiddersHash` was in old spec, and list was added later. So I guess it is just outdated.

I propose not to implement it now and wait for @GeorgeFlerovsky return to ask  if we missed something.

src/HydraAuction/Tx/Common.hs
uhbif19: You are meaning Commons -> StandingBid dep? Why? I am talking about moving all three functions there and there is no other deps from Commons as far as I see.

nazrhom: Because `Escrow` uses those functions, but `StandingBid` imports `Escrow`

src/HydraAuction/Tx/Escrow.hs
nazrhom: Perfect, I have added that anyway. I can keep it in a separate PR, and when @GeorgeFlerovsky is back we can disciuss

README.md
nazrhom: I think it is correct, as we are making a list of things 🤷 
I use grammarly, and if it does not get angry assume wording is ok, but I am not a native speaker either 😅

src/HydraAuction/Fixture.hs
nazrhom: It's just that both the CLI and the tests need it often

src/HydraAuction/Tx/Common.hs
uhbif19: You mean `toForgeStateToken`? Hm. Maybe we can just it to other module? Like `Tx.Commons`.

Or we can do separate module like `StandingBidCommons` or something like that.

README.md
uhbif19: Okay, Grammarly seems to be okay with that.

uhbif19 (approved):

Issue #178, opened 2023-03-09
Separate all domain-independent utilities to separate modules and cabal lib


uhbif19:
- [x] New cabal lib `hydra-auction-utils`
- [x] Move `Fixture`
- [x] Move extras
- [x] Separate utils from `Tx.Common`
     - [x] AutoCreateTx and Utxo filtering
     - [x] Module for `currentTimeX` (also separate `POSIXTime <$> currentTimeMilliseconds` into function)
     - [x] Build tx stuff, like `mintedTokens` and `mkInlinedDatumScriptWitness`
     - [x] Runner specific things like `addressAndKeys` and `actorTipUtxo` (or intoduce `MonadActor`)
- [x] Separate utils from `OnChain.Common`
- [x] Separate `Natural` from `Types`
- [x] Separate common stuff from `app/`, for example pretty-printers and parsers

Issue #125, opened 2023-02-20
Add support for Hydra in off-chain util functions


uhbif19:
Probably we need to add Runner param for hydra-node. I think in end we need to have most of tx submitting functions work on both L1 or L2 dependent on this param.

* Support submitting custom Tx on hydra
* Support query UTxOs from Hydra-node
     * Decide on snapshots and ouptut parsing
     * Add some kind of polymorphism to query functions I guess
* Add L2 version of E2E Ledger test

PR #215, opened 2023-03-24
Separate utils lib and add monadic interfaces to it

uhbif19:
Separating monads as part of #125 
We do not really need ad-hoc polymorphism in that cases,
but those functions should be reused for L1/L2 cases anyway,
so it is nice way to group them.

Also some work on #178

### DIFF
--- a/src-utils/HydraAuctionUtils/Monads.hs
+++ b/src-utils/HydraAuctionUtils/Monads.hs
@@ -106,5 +106,5 @@ submitAndAwaitTx tx = do
    For example we could provide QueryUtxo but not SubmitTx
    for Delegate server clients.
 -}
-type MonadCardanoClient m =
-  (MonadQueryUtxo m, MonadNetworkId m, MonadSubmitTx m)
+class (MonadQueryUtxo m, MonadNetworkId m, MonadSubmitTx m) => MonadCardanoClient m
+instance (MonadQueryUtxo m, MonadNetworkId m, MonadSubmitTx m) => MonadCardanoClient m
--- a/src-utils/HydraAuctionUtils/Monads.hs
+++ b/src-utils/HydraAuctionUtils/Monads.hs
@@ -1,0 +1,76 @@
+module HydraAuctionUtils.Monads (
+  UtxoQuery (..),
+  MonadTrace (..),
+  MonadSubmitTx (..),
+  MonadQueryUtxo (..),
+  MonadNetworkId (..),
+  MonadCardanoClient,
+  logMsg,
+  submitAndAwaitTx,
+) where
+
+-- Prelude imports
+import Prelude
+
+-- Cardano imports
+import Cardano.Api.UTxO qualified as UTxO
+import Hydra.Cardano.Api (
+  Address,
+  NetworkId,
+  ShelleyAddr,
+  Tx,
+  TxIn,
+  getTxId,
+  txBody,
+ )
+
+-- HydraAuctionUtisl imports
+import HydraAuctionUtils.Fixture (Actor)
+
+-- MonadQueryUtxo
+
+data UtxoQuery
+  = ByActor Actor
+  | ByAddress (Address ShelleyAddr)
+  | ByTxIns [TxIn]
+
+class Monad m => MonadQueryUtxo m where
+  queryUtxo :: UtxoQuery -> m UTxO.UTxO
+
+-- MonadNetworkId
+
+class Monad m => MonadNetworkId m where
+  askNetworkId :: m NetworkId
+
+-- MonadTrace
+
+class Monad m => MonadTrace m where
+  type TracerMessage m = mt | mt -> m
+  stringToMessage :: String -> TracerMessage m
+  traceMessage :: TracerMessage m -> m ()
+
+logMsg :: MonadTrace m => String -> m ()
+logMsg = traceMessage . stringToMessage
+
+-- MonadSubmitTx
+
+class Monad m => MonadSubmitTx m where
+  submitTx :: Tx -> m ()
+  awaitTx :: Tx -> m ()
+
+submitAndAwaitTx :: (MonadSubmitTx m, MonadTrace m) => Tx -> m ()
+submitAndAwaitTx tx = do
+  submitTx tx
+  logMsg "Submited"
+  awaitTx tx
+  logMsg $ "Created Tx id: " <> show (getTxId $ txBody tx)
+
+-- Complex constraint synonims
+
+{- Why split monads at all?
+   I think in some cases we may not have them for same monad.
+   For example we could provide QueryUtxo but not SubmitTx
+   for Delegate server clients.
+-}
+type MonadCardanoClient m =
+  (MonadQueryUtxo m, MonadNetworkId m, MonadSubmitTx m)
--- a/src-utils/HydraAuctionUtils/Monads.hs
+++ b/src-utils/HydraAuctionUtils/Monads.hs
@@ -7,6 +7,7 @@ module HydraAuctionUtils.Monads (
   MonadCardanoClient,
   logMsg,
   submitAndAwaitTx,
+  fromPlutusAddressInMonad,
 ) where
 
 -- Prelude imports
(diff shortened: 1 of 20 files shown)


src/HydraAuction/Runner.hs
```diff
@@ -69,8 +88,8 @@ data ExecutionContext = MkExecutionContext
   , actor :: !Actor
   }
 
-{- | Hydra computation executor. Note that @Runner@ is
- de facto `ReaderT ExecutionContext IO`.
+{- | HydraAuction specific L1 computation executor.
```
nazrhom, line 91: I like these changes, maybe we could rename runner to something like `CardanoRunner`, just to make it clear this is operating on L1

uhbif19: Yep, I thought about this too.

nazrhom:
There are still a few build errors and conflicts, but the idea LGTM

nazrhom (approved):

src-utils/HydraAuctionUtils/Monads.hs
```diff
@@ -0,0 +1,110 @@
...
+addressAndKeysForActor actor = do
+  networkId' <- askNetworkId
+  (actorVk, actorSk) <- liftIO $ keysFor actor
+  let actorAddress = buildAddress actorVk networkId'
+  pure (actorAddress, actorVk, actorSk)
+
+-- MonadTrace
+
+class Monad m => MonadTrace m where
+  type TracerMessage m = mt | mt -> m
```
MangoIV, line 82: this means that if you know the trace message you always know the Monad? does that make sense? I don't see that e.g. for `String`s there can only be one valid `Monad` to trace in

MangoIV: by chance, I wrote a similar thing for my current issue: 
```haskell
class MonadTracer t m | m -> t where
  trace :: t -> m ()

-- | a transformer that provides a tracer
newtype TracerT t m a = TracerT {runTracerT :: ReaderT (Tracer m t) m a}
  deriving newtype
    ( Functor
    , Applicative
    , Monad
    , MonadReader (Tracer m t)
    , MonadIO
    )

-- these could be derived in a normal deriving clause but I feel better specifying
-- the context myself
deriving newtype instance MonadWriter w m => MonadWriter w (TracerT t m)
deriving newtype instance MonadState s m => MonadState s (TracerT t m)

mapTracerT :: (t1 -> t2) -> TracerT t1 m a -> TracerT t2 m a
mapTracerT f = TracerT . withReaderT (contramap f) . runTracerT

runWithTracer :: (forall b. n b -> m b) -> Tracer n t -> TracerT t m a -> m a
runWithTracer natTrans tracer = flip runReaderT (natTracer natTrans tracer) . runTracerT

runWithTracer' :: Tracer m t -> TracerT t m a -> m a
runWithTracer' = runWithTracer id

askTracer :: forall t m. Monad m => TracerT t m (Tracer m t)
askTracer = ask @(Tracer m t) 
```

src-utils/HydraAuctionUtils/Monads.hs
```diff
@@ -0,0 +1,110 @@
...
+
+-- Complex constraint synonims
+
+{- Why split monads at all?
+   I think in some cases we may not have them for same monad.
+   For example we could provide QueryUtxo but not SubmitTx
+   for Delegate server clients.
+-}
+type MonadCardanoClient m =
+  (MonadQueryUtxo m, MonadNetworkId m, MonadSubmitTx m)
```
MangoIV:
```suggestion
class (MonadQueryUtxo m, MonadNetworkId m, MonadSubmitTx m) => MonadCardanoClient m
instance (MonadQueryUtxo m, MonadNetworkId m, MonadSubmitTx m) => MonadCardanoClient m
```

I think this is preferred because it circumvents some issues (e.g. with `-XQuantifiedConstraints`) and it can be partially applied

src-utils/HydraAuctionUtils/Monads.hs
```diff
@@ -0,0 +1,110 @@
...
+import HydraAuctionUtils.Fixture (Actor, keysFor)
+
+-- MonadQueryUtxo
+
+data UtxoQuery
+  = ByAddress (Address ShelleyAddr)
+  | ByTxIns [TxIn]
+
+class Monad m => MonadQueryUtxo m where
+  queryUtxo :: UtxoQuery -> m UTxO.UTxO
```
MangoIV, line 50: I really like these :)

src-utils/HydraAuctionUtils/Monads.hs
```diff
@@ -0,0 +1,110 @@
...
+class Monad m => MonadSubmitTx m where
+  submitTx :: Tx -> m ()
+  awaitTx :: Tx -> m ()
+
+submitAndAwaitTx :: (MonadSubmitTx m, MonadTrace m) => Tx -> m ()
+submitAndAwaitTx tx = do
+  submitTx tx
+  logMsg "Submited"
+  awaitTx tx
+  logMsg $ "Created Tx id: " <> show (getTxId $ txBody tx)
```
MangoIV, line 100: this works fine if all you want to trace are `String`s but as soon as it is some other type (e.g. a structured log type) you will have `TracerMesage m ~ MyLogtype` in your context, I think it's arguably nicer to have `MonadTrace MyLogType m`

MangoIV (commented):
Looks largely good to me, although I would do a couple of things differently as noted. :)

src-utils/HydraAuctionUtils/Monads.hs
MangoIV: (yes `TracerT` is not really that relevant, except for their implementation of `MonadTrace` is I guess similar to the one of your ReaderT IO newtype)

uhbif19: > does that make sense? I don't see that e.g. for Strings there can only be one valid Monad to trace in

In all places we use tracer they always define custom log datatypes, like HydraAuctionLog or EndToEndLog. Do not know why so, but this is current state of affairs.

> TracerT

I think we wont need a separate transformer, cuz all Runners have custom context which may have tracer.

src-utils/HydraAuctionUtils/Monads.hs
uhbif19: `logMsg` is made specifically to emit strings. There is fundep type for structured log. So I guess no need for such constraints.

So for now typeclass assumes that String can be embedded to structured log type. This may be strange but makes `logMsg`, used everywhere in our code.

MangoIV: oh, that's not correct, because it's injective, but yeah, it kinda shows the issue, you want to be able to trace multiple log types, the one-to-one correspondence is a bit odd, I think. okay, I think I have to think about it again

src-utils/HydraAuctionUtils/Monads.hs
MangoIV: > I think we wont need a separate transformer, cuz all Runners have custom context which may have tracer.

my thing uses this so that's that

MangoIV: > In all places we use tracer they always define custom log datatypes, like HydraAuctionLog or EndToEndLog. Do not know why so, but this is current state of affairs.

that's not the point though, you really will not be able to say 
`instance MonadTrace SomeMonad` and `instance MonadTrace SomeOtherMonad` where the log type is not fixed, you will always have to choose one specific log type, so that completely defeats the purpose because it can never be polymorphic.

MangoIV: if you have a log type, you will have a fixed Monad, in that case you can just monomorphise...

src-utils/HydraAuctionUtils/Monads.hs
MangoIV: > So for now typeclass assumes that String can be embedded to structured log type. 

it will require parsing the strings into some structured log type at some point (if we want to go that route, but I don't think so, in that case we just never use a structured log and instead just trace directly to stdout)

src-utils/HydraAuctionUtils/Monads.hs
uhbif19: > so that completely defeats the purpose

Which purpose? My purpose is to make log and use tracer on different monads. Because my functions are polymorphic on monads.

Also I remember why I added fundep: otherwise `logMsg` wont work. There may be other ways to do it but I think we can easily change that interface later. You can open issue if you wanna discuss that further.

src-utils/HydraAuctionUtils/Monads.hs
uhbif19: > because it's injective

What is injective?

> it will require parsing the strings into some structured log type at some point

No, we just pushing strings in datatype as you may see. Any such log place could be refactored to structured log later.

MangoIV: > What is injective?

the type family, what you added is an injectivity annotation

MangoIV: > Any such log place could be refactored to structured log later.

no, I don't think so.

Issue #203, opened 2023-03-16
Add bidder authentication in NewBid transactions


uhbif19:
Current plan is to include bidder signature in Tx datum and check it on chain.

### Possible signing solutions

We probably would need both signatures. Bidder signature should be checked on chain, delegate signature is required to spend collateral (see #193). First signature only should be accessible by on-chain code, while second can only be regular Cardano Tx signature

There is multiple ways to do signing in presence of delegate server:

1. Frontend sign datum and send it to delegate server. Delegate server creates, signs (optionally) and submits Tx.
2. Create and sign whole Tx by Frontend. Delegate server only submits it to L2.
3. Create and sign whole Tx by Frontend. Delegate server parses Tx, signs it and submits to L2.

I think all of them are possible.

Signed datum actually can be checked on chain, because three is `verifySignature` Plutus primitive and datum could be serialized by hand (see: https://app.slack.com/client/T01UN827FQV/C028AD4LJGY).

Parsing by delegate server looks to be possible using of `SerialiseCBOR Tx` instance.

### Selected signing solution

Solution 1 costs a lot on chain, because we need to re-serialize part of datum and serializing integer  works bad on chain. See #220 
Solution 2 does not provide delegate signature.
Solution 3 is bad because requires signing by bidder. In our model we presume that in production this should be done with user wallet, so this would be slow, while Tx input on L2 may be invalidated quickly. Also this requires Tx serialization (which only works with CBOR).

**UPD:** Selected solution is 1, as long as transactions fit within blockchain constraints. We will deal with tx optimization later.
@GeorgeFlerovsky 

That means that last off-chain part of issue is blocked by #158 and #175

### Solution parts:

- [x] Select signing scheme
- [ ] Check that bidder signed bid and ensure delegate server/frontend do those signing

nazrhom:
@uhbif19 I thought the decision here was to parametrise the auction terms by a list of accepted bidders (cfr. https://github.com/mlabs-haskell/hydra-auction/blob/staging/doc/adr/2023-01-09_003-open_vs_closed_auctions.md)

nazrhom:
Current implementation plan (requires #158)

- add signers field to autoCreateTx
- create Tx.StandingBid.newBidL2 which does not use collateral and returns tx
- create newBidL2 in cli Actions which uses new function to generate tx, serialise it to CBOR and submits it to delegate server
- update delegate server to receive CBOR encoded tx in new bid
[...]

PR #216, opened 2023-03-24
Refactor Tx body creation and signing

nazrhom:


nazrhom:
Separating this will be useful for the second part of #203

src/HydraAuction/Tx/Escrow.hs
```diff
@@ -136,6 +136,7 @@ announceAuction terms = do
     autoSubmitAndAwaitTx $
       AutoCreateParams
         { authoredUtxos = [(sellerSk, utxoWithLotNFT <> sellerMoneyUtxo)]
+        , signers = [sellerSk]
```
uhbif19: Why add extra signer here, if it is already in `authoredUtxos`?

src/HydraAuction/Tx/Common.hs
```diff
@@ -318,7 +320,18 @@ toSlotNo ptime = do
   either (error . show) return $ slotFromUTCTime timeHandle utcTime
 
 autoCreateTx :: AutoCreateParams -> Runner Tx
-autoCreateTx (AutoCreateParams {..}) = do
+autoCreateTx params = do
+  body <- autoCreateTxBody params
+  pure $ signTx (signers params) body
```
uhbif19: It should use `authoredUtxos` signers as well.

uhbif19 (changes requested):

src/HydraAuction/Tx/Common.hs
nazrhom: Can you explain how we are using `authoredUtxos`? We use it to fill `TxExtraKeyWitnesses`, but that requires a VKey, we are passing in an SKey and every user will only have access to their own SKey.

I can not find the official docs on `TxExtraKeyWitnesses`, but my understanding is that you specify extra VKeys that will need to sign the tx. For example, the bidder could construct a Tx which spends the collateral of the delegate, and would need to add the delegate VKey as extra key witness? Maybe I am getting it wrong though, could you also clarify what that field means in the tx then?

uhbif19: > Can you explain how we are using authoredUtxos?

It is bunch of UTxOs paired with keys to sign them. That API makes clear which UTxO uses which keys and prevents error on missing signing witness or signing.

Not sure that I fully understand your question thought.

src/HydraAuction/Tx/Common.hs
```diff
@@ -294,6 +294,8 @@ scriptUtxos script terms = do
 
 data AutoCreateParams = AutoCreateParams
   { authoredUtxos :: [(SigningKey PaymentKey, UTxO)]
+  , signers :: [SigningKey PaymentKey]
```
uhbif19: Can you name it `additionalSigners` or something like that? Cuz currently is not clear that there is an other thing that signs.

Also we can rename `authoredUtxos` into `signedUtxos`.

uhbif19 (approved):

PR #217, opened 2023-03-27
[feat] initial draft of delegate server

### LINKED ISSUE #175: Implement real network interface to be used by Frontend for Delegate server
See for library to use plan: https://github.com/mlabs-haskell/hydra-auction/issues/141#issuecomment-1473007419

Blocked by #140 
Blocked by #141

MangoIV:
- add Delegate Server and types
- rewrite the Main to work with a proper server
- some changes to handle serialisation


delegate-app/HydraAuction/Delegate/Server.hs
```diff
@@ -0,0 +1,119 @@
...
+data DelegateError = FrontendNoParse
+  deriving stock (Eq, Show, Generic)
+
+instance Pretty DelegateError where
+  pretty = \case
+    FrontendNoParse -> "Could not parse the input provided by the frontend"
+
+type DelegateTracer = TracerT DelegateServerLog
+
+class MonadTracer t m | m -> t where
```
uhbif19: Please separate transformer/monad stuff to separate file.

Also I do not understand what is need for this. Like consumer is just IO thread, it could just log into Tracer constant.

delegate-app/HydraAuction/Delegate/Server.hs
```diff
@@ -0,0 +1,119 @@
...
+instance Pretty DelegateServerLog where
+  pretty = \case
+    Started port -> "Started Server at Port" <> pretty (show port)
+    FrontendConnected -> "Frontend connected to Server"
+    DelegateOutput out -> "Delegate output" <> line <> viaShow out
+    FrontendInput inp -> "Frontend input" <> line <> viaShow inp
+    DelegateError err -> "Delegate error" <> line <> pretty err
+
+-- FIXME: needs actual error
+data DelegateError = FrontendNoParse
```
uhbif19: What would we do with error datatype?

Client should get error by async response anyway. So I guess this should be fit in regular `delegateStep` logic.

delegate-app/HydraAuction/Delegate/Server.hs
```diff
@@ -0,0 +1,119 @@
...
+  = Started PortNumber
+  | FrontendConnected
+  | FrontendInput FrontendRequest
+  | DelegateOutput DelegateResponse
+  | DelegateError DelegateError
+  deriving stock (Eq, Show, Generic)
+
+instance Pretty DelegateServerLog where
+  pretty = \case
+    Started port -> "Started Server at Port" <> pretty (show port)
```
uhbif19: All other logs just use show so this may look strange in CLI.

delegate-app/HydraAuction/Delegate/Server.hs
```diff
@@ -0,0 +1,119 @@
...
+  { dlgt'host :: IP
+  -- ^ the host of the delegate server
+  , dlgt'port :: PortNumber
+  -- ^ the port number the delegate server receives input at
+  , dlgt'tick :: Int
+  -- ^ the amount of milliseconds, the thread should wait
+  }
+
+data DelegateServerLog
+  = Started PortNumber
```
uhbif19, line 56: Cool.

delegate-app/Main.hs
```diff
@@ -1,49 +1,133 @@
...
-  where
-    go thisIsFirstRun = do
-      mEvent <- eventsProducer thisIsFirstRun
-      consumer (DelegateEvent <$> mEvent)
-      liftIO $ threadDelay delegateTick
-      go False
+  port <- lookupEnv "port"
+
+  -- FIXME: do we need this?
+  let host :: IP = "127.0.0.1"
```
uhbif19: I think while style is better cuz it makes clear that no monadic side-effects are happening.

delegate-app/Main.hs
```diff
@@ -1,49 +1,133 @@
...
+  Int ->
+  ServerAppT m
+delegateServerApp tick pending = do
+  connection <- liftIO $ acceptRequest pending
+  trace FrontendConnected
+  -- FIXME: we will probably have to do this in parallel, but how can we enqueue in the same
+  -- state? probably have to remodel this
+  execDelegateRunnerT $ flip evalStateT True $ mkRunner tick connection
+
+-- FIXME: remove concrete monads and make this tagless final (needs MonadDelegate)
```
uhbif19: Why remove concrete monads? This should just be called from main, no?

delegate-app/Main.hs
```diff
@@ -1,49 +1,133 @@
...
+
+  liftIO $ threadDelay tick
+  put False
+
+runDelegateServer ::
+  DelegateServerConfig ->
+  TracerT DelegateServerLog IO ()
+runDelegateServer conf = do
+  trace (Started $ dlgt'port conf)
+  let fallback :: Application
```
uhbif19: What does this code do?

delegate-app/Main.hs
```diff
@@ -1,49 +1,133 @@
...
-        "Delegate responses for input: " <> show delegateResponse
-  Nothing -> return ()
+delegateServerApp ::
+  (MonadIO m, MonadTracer DelegateServerLog m) =>
+  Int ->
+  ServerAppT m
+delegateServerApp tick pending = do
+  connection <- liftIO $ acceptRequest pending
+  trace FrontendConnected
+  -- FIXME: we will probably have to do this in parallel, but how can we enqueue in the same
```
uhbif19: Yes, we discussed that before and my initial version had StateT tood. And then we dropped state so no such tricks are needed anymore.

Whole point was that we can just do dumb stateless producer threads in `IO` monad, and leave all custom monadic stuff to `delegateStep`.

Also with parallel threads we can just have one-shot thread which posts `Started` event and drop `isFirstRun` fully.

uhbif19 (changes requested):

delegate-app/Main.hs
uhbif19: Like it is some HTTP-specific stuff. Hydra just uses `runServer` and that seems enough.

Hydra code is probably over complicated over what we need.

delegate-app/HydraAuction/Delegate/Server.hs
MangoIV: yes, I just haven't done it yet, thank you

delegate-app/HydraAuction/Delegate/Server.hs
MangoIV: well those are the errors that are emitted by the thing that calls `execDelegateRunner`, so we cannot emit them from inside the monad, e.g. this specific appears if we cannot parse the JSON that the frontend inputs

delegate-app/HydraAuction/Delegate/Server.hs
MangoIV: this is what IOG uses largely, I think it's quite nice because we can just in a modular way structure our logs and adjust the error messages whenever we want to. I don't see why it would look strange in CLI, it would just be formatted a bit differently, I think it's nicer than adhoc strings in the code.

MangoIV: this also serves as documentation

delegate-app/Main.hs
MangoIV: I think this was mainly because it was easier to read the tagless final form than having concrete monads in this case. also we don't lose anything (perhaps performance but I don't even think that's the case here)

delegate-app/Main.hs
MangoIV: in the docs it explicitly said you should not use `runServer`, that's why I didn't, this basically offers a fallback if the frontend tries to communicate via HTTP instead of websockets

delegate-app/Main.hs
MangoIV: > Whole point was that we can just do dumb stateless producer threads in IO monad, and leave all custom monadic stuff to delegateStep.

I mostly put it there because having `StateT` shows better that we pass some state that is changed later then updating it by hand. but if you want I can change it back

delegate-app/HydraAuction/Delegate/Server.hs
MangoIV:
```suggestion
    Started port -> "Started Server at Port" <> viaShow port
```

delegate-app/Main.hs
MangoIV: https://hackage.haskell.org/package/websockets-0.12.7.3/docs/Network-WebSockets.html#v:runServer

MangoIV:
I think a possible solution for the concurrency issue is having a `TChan` that the `DelegateRunner` knows and then every `tick` read of the `TChan` all FrontendRequests, then do stuff, after that write to a `broadCastTChan` which will then be cloned by the clients as soon as they're spawned with `dupTChan`

MangoIV:
I will implement that tomorrow then

delegate-app/HydraAuction/Delegate/Server.hs
uhbif19: Yes. But still I not see need for Either. This code should just return some answer like `WrongInput` to client and add nothing to inputs queue. That is that Hydra does in such case AFAIU.

delegate-app/HydraAuction/Delegate/Server.hs
uhbif19: > this is what IOG uses largely, I think it's quite nice because we can just in a modular way structure our logs and adjust the error messages whenever we want to. ..., I think it's nicer than adhoc strings in the code.

Of course, I completely agree.

> I don't see why it would look strange in CLI, it would just be formatted a bit differently, 

Because different formatting stands out when reading logs. Anyway this does not really matter, we could add some prefix or something else if that would be a problem.

delegate-app/Main.hs
uhbif19: *`where` style

MangoIV:
Inwards it can be a TQueue, I think. Outwards I'll use TChan

delegate-app/Main.hs
uhbif19: >  it was easier to read the tagless final form than having concrete monads

Why? `MonadDelegate` does not seem to have any laws or any logic at all.

> also we don't lose anything

Introduction of new interface adds code and necessity to read it.

delegate-app/Main.hs
uhbif19: > this basically offers a fallback if the frontend tries to communicate via HTTP instead of websockets

But it should not do that. Our frontend is not browser.

> Glue for using this package with real servers is provided by:

I think they are talking about integration with HTTP server which we do not do anyway.

Their points: about DDoS is not relevant for our scope and logging is already covered in your PR.

delegate-app/Main.hs
uhbif19: > but if you want I can change it back

We do not need that state after #218 anymore anyway.

delegate-app/Main.hs
nazrhom: We can parse these as args (we will need it also for hydra node later). We can use a similar style as we use in the CLI

nazrhom:
Hey @MangoIV have a look at [this pr](https://github.com/mlabs-haskell/hydra-auction/pull/218) I have already added a `Chan` to queue events in, and `delegateStep` can consume that directly.

Also - this might have been already discussed - but I don't think we should pass a tick here. Ideally we can listen for websocket connections and spawn a thread thread to handle each incoming request.

delegate-app/HydraAuction/Delegate/Server.hs
MangoIV: Well yeah, this doesn't get enqueued anywhere

delegate-app/Main.hs
MangoIV: I will change it to `runServer`

delegate-app/Main.hs
MangoIV: I removed the state as part of my changes wrt channels, we can just write a Started event to the incoming channel at the very beginning

delegate-app/HydraAuction/Delegate/Server.hs
MangoIV: Reminder to self to respond with this error if there occurs a failure in encoding

delegate-app/Main.hs
MangoIV: > We can parse these as args (we will need it also for hydra node later). We can use a similar style as we use in the CLI

Is this in scope of this PR?

hydra-auction.cabal
```diff
@@ -183,9 +183,23 @@ executable hydra-auction
...
   import:         common-lang
   main-is:        Main.hs
+  other-modules:
+    HydraAuction.Delegate.Server
+    HydraAuction.Delegate.Tracing
+
   hs-source-dirs: delegate-app
   ghc-options:    -threaded -rtsopts
-  build-depends:  hydra-auction
+  build-depends:
```
MangoIV, line 184: I removed these, but forgot to push

src/HydraAuction/Types.hs
```diff
@@ -85,11 +88,7 @@ naturalToInt (Natural i) = i
 PlutusTx.makeLift ''Natural
 
 instance FromJSON Natural where
-  parseJSON x = do
-    int <- parseJSON x
-    case intToNatural int of
-      Just nat -> return nat
-      Nothing -> fail "Integer is not natural"
+  parseJSON = maybe (fail "Integer is not natural") return . intToNatural <=< parseJSON
```
uhbif19, line 93: What is wrong with do notation?

delegate-app/HydraAuction/Delegate/Tracing.hs
```diff
@@ -0,0 +1,73 @@
+{-# LANGUAGE FunctionalDependencies #-}
```
uhbif19: That should belong to hydra-auction-utils dir.

uhbif19: Cuz is not domain-specific.

delegate-app/HydraAuction/Delegate/Tracing.hs
```diff
@@ -0,0 +1,73 @@
...
+import Control.Monad.State (MonadState)
+import Control.Monad.Writer (MonadWriter)
+import Control.Tracer (Tracer, natTracer, traceWith)
+
+{- | a monad that one can trace in
+   the typeclass takes as argument the
+   type that serves as the trace and the monad
+   that you can trace in
+-}
+class MonadTracer t m | m -> t where
```
uhbif19: We already have one such interface. So you need integrate this one into existing one. Like make it superclass.

And I am still not understanding why fundeps are better here. Type family have one obvious advantage - it does not require to write second variable. What is advantage of fundeps?

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
 main :: IO ()
 main = do
-  -- FIXME: cover either case. It probably should not be transformer.
-  _ <- execDelegateRunnerT $ go True
-  return ()
+  port <- lookupEnv "PORT"
+
+  let conf :: DelegateServerConfig
+      conf =
+        DelegateServerConfig
```
uhbif19, line 233: Config stuff is cool.

delegate-app/HydraAuction/Delegate/Server.hs
```diff
@@ -0,0 +1,78 @@
...
+-- Cardano imports
+import Hydra.Network (IP, PortNumber)
+
+-- Hydra auction imports
+import HydraAuction.Delegate.Interface (DelegateResponse, FrontendRequest)
+import HydraAuction.Delegate.Tracing (TracerT)
+
+-- | The config for the delegate server
+data DelegateServerConfig = DelegateServerConfig
+  { dlgt'host :: IP
```
uhbif19: That is style is different for all other repo. Style should be same everythere.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+  events <- liftIO . atomically $ flushTQueue reqq
+  traverse_ (delegateStep >=> encodeBroadcast) events
+
+  liftIO $ threadDelay tick
+
+-- | run a delegate server
+runDelegateServer ::
+  -- | the configuration of the delegate server
+  DelegateServerConfig ->
+  TracerT DelegateServerLog IO ()
```
uhbif19: You not using your `DelegateTracerT` thing here as in other places. While I do not see why `TracerT DelegateServerLog` is worse, but still.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+   accordingly
+-}
+mkRunner ::
+  -- | the time in milliseconds that the runner sleeps between acts
+  Int ->
+  -- | the queue of incoming messages
+  TQueue DelegateInput ->
+  -- | the broadcast queue of outgoing messages (write only)
+  TChan DelegateResponse ->
+  DelegateRunnerT (DelegateTracerT IO) ()
```
uhbif19: I think that term `Runner` is already used as name for monad pattern we use, not for actual code block.

uhbif19 (changes requested):
Still think `mbQueueAuctionPhases` is not best way to separate function.

src/HydraAuction/Types.hs
MangoIV: Idk, it was just something random I did because I like the fish and maybe :D

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+-- FIXME: we need to abort at some point but this doesn't seem
```
uhbif19, line 144: Also we do not handle Ctrl-C reaction and thread finalization stuff at all. I am not aware what is proper way to deal with Ctrl-C.

delegate-app/Main.hs
MangoIV: I think I don't understand this.

delegate-app/Main.hs
MangoIV: I'm using it in Main

MangoIV: But I can remove it if you wish.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+  DelegateTracerT,
+  ServerAppT,
+ )
+import HydraAuction.Delegate.Tracing (MonadTracer (trace), TracerT, askTracer, runWithTracer')
+
+{- | per websocket, a thread gets spawned which then
+   subscribes to the broadcast channel and sends to
+   the incoming channel of the delegate runner
+-}
+delegateServerApp ::
```
uhbif19, line 87: I do not understand naming of this function.

delegate-app/HydraAuction/Delegate/Tracing.hs
MangoIV: The obvious advantage is that you don't need to specify an equality constraint if you want a specific type to trace with.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+
+  delegateInputChan <- liftIO . atomically $ do
+    q <- newTQueue
+    writeTQueue q (DelegateEvent Start)
+    pure q
+  delegateBroadCastChan <- liftIO . atomically $ newTChan
+
+  _ <-
+    liftIO . forkIO $
+      runServer (show $ dlgt'host conf) (fromIntegral $ dlgt'port conf) $
```
uhbif19: > fromIntegral

Why not just store it in that type already in config?

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+        runWithTracer' tracer . delegateServerApp (dlgt'ping conf) delegateInputChan delegateBroadCastChan
+
+  execDelegateRunnerT $ mkRunner (dlgt'tick conf) delegateInputChan delegateBroadCastChan
```
uhbif19: This still looks like it only works once.

uhbif19 (changes requested):

delegate-app/Main.hs
MangoIV: What do you mean it works only once? This replies to requests forever.

delegate-app/Main.hs
MangoIV: Because it's basically untyped. The proper server I had before took proper types. This doesn't. If you wish I can change it. Be aware that we don't validate IPs then anymore (which we do now by parsing IPs)

delegate-app/Main.hs
MangoIV: It's a ServerApp and specifically the serverApp that is run by the delegate.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+  trace FrontendConnected
+  tracer <- askTracer
+
+  liftIO $ do
+    broadcast' <- atomically $ dupTChan broadcast
+
+    let encodeSend :: forall a. ToJSON a => a -> IO ()
+        encodeSend = sendTextData connection . encode
+
+        receive = forever $
```
uhbif19, line 108: Please move this in `where` blocks or separate functions.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+          runWithTracer' tracer $ do
+            inp <- liftIO $ receiveData connection
+            case eitherDecode @FrontendRequest inp of
+              Left msg -> do
+                let err = FrontendNoParse msg
+                trace (DelegateError err)
```
uhbif19, line 114: I do not think that this should be named `DelegateError`. This is error on client side, for delegate it is just normal event.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+  DelegateRunnerT (DelegateTracerT IO) ()
+-- FIXME: we need to abort at some point but this doesn't seem
+-- to be implemented yet so we just go on
+mkRunner tick reqq broadcast = forever $ do
+  let encodeBroadcast responses = do
+        liftIO . atomically $ traverse_ (writeTChan broadcast) responses
+        lift $ traverse_ (trace . DelegateOutput) responses
+
+  events <- liftIO . atomically $ flushTQueue reqq
+  traverse_ (delegateStep >=> encodeBroadcast) events
```
uhbif19, line 152: Your PR does not address broadcast/per-client output distinction problem we discussed earlier.

MangoIV: I wasn't aware that was in scope.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+  TChan DelegateResponse ->
+  DelegateRunnerT (DelegateTracerT IO) ()
+-- FIXME: we need to abort at some point but this doesn't seem
+-- to be implemented yet so we just go on
+mkRunner tick reqq broadcast = forever $ do
+  let encodeBroadcast responses = do
+        liftIO . atomically $ traverse_ (writeTChan broadcast) responses
+        lift $ traverse_ (trace . DelegateOutput) responses
+
+  events <- liftIO . atomically $ flushTQueue reqq
```
uhbif19, line 151: I guess you right that there are the cases when this works better, but this does not "solve" queue overload issue. I think that PR is right place to document that in `off_chain_spec.md`. Or I can do it in my PR.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+  Int ->
+  -- | the queue of incoming messages
+  TQueue DelegateInput ->
+  -- | the broadcast queue of outgoing messages (write only)
+  TChan DelegateResponse ->
+  DelegateRunnerT (DelegateTracerT IO) ()
+-- FIXME: we need to abort at some point but this doesn't seem
+-- to be implemented yet so we just go on
+mkRunner tick reqq broadcast = forever $ do
+  let encodeBroadcast responses = do
```
uhbif19, line 147: Please use `where`.

delegate-app/Main.hs
MangoIV: The delegate runner t doesn't see this. This follows the pattern of naming a constr the same as the type that it takes that you use in the project. Err is of type DelegateError.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+  -- | the time in milliseconds that the runner sleeps between acts
+  Int ->
+  -- | the queue of incoming messages
+  TQueue DelegateInput ->
+  -- | the broadcast queue of outgoing messages (write only)
+  TChan DelegateResponse ->
+  DelegateRunnerT (DelegateTracerT IO) ()
+-- FIXME: we need to abort at some point but this doesn't seem
+-- to be implemented yet so we just go on
+mkRunner tick reqq broadcast = forever $ do
```
uhbif19, line 146:
```suggestion
mkRunner tick delegateInputChan broadcast = forever $ do
```

delegate-app/Main.hs
uhbif19: Yeah, I see, I men you use it in other places but not this one => not consitent.

delegate-app/Main.hs
MangoIV: Yes. It only solves it if we can assume that the tick is the time consuming part.

delegate-app/Main.hs
uhbif19: Also, why Chan if it is queue.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+delegateServerApp pingSecs reqq broadcast pending = do
+  connection <- liftIO $ acceptRequest pending
+
+  trace FrontendConnected
+  tracer <- askTracer
+
+  liftIO $ do
+    broadcast' <- atomically $ dupTChan broadcast
+
+    let encodeSend :: forall a. ToJSON a => a -> IO ()
```
uhbif19, line 104:
```suggestion
    let encodeAndSend :: forall a. ToJSON a => a -> IO ()
```

delegate-app/Main.hs
MangoIV: A TChan is just a Queue with a different implementation. I named it reqq for request queue.

delegate-app/Main.hs
MangoIV: Oversight. Thank you.

delegate-app/Main.hs
MangoIV: These use parameters that are brought in scope in the do block. These are not in scope in a where. They would have to be functions that take two parameters which imho is unnecessarily verbose.

delegate-app/Main.hs
```diff
@@ -1,49 +1,179 @@
...
+              Right req -> do
+                trace $ FrontendInput req
+                liftIO . atomically . writeTQueue reqq $ FrontendRequest req
+
+        send =
+          forever $
+            atomically (readTChan broadcast')
+              >>= encodeSend
+
+    withPingThread connection pingSecs (pure ()) $ race_ receive send -- TODO: can we do this
```
uhbif19: > -- Many (but not all) proxies have a 60 second default timeout, so based on that
> -- sending a ping every 30 seconds is a good idea.

Please add code comment on why code uses this ping stuff.

delegate-app/Main.hs
uhbif19: I was wrong, missed forever.

delegate-app/Main.hs
MangoIV: I don't understand how this would improve the function or add additional semantics to the name.

delegate-app/Main.hs
uhbif19: Yes, you can just move that consturctor- from `DelegateError` type.

delegate-app/Main.hs
uhbif19: Yes. I know. This makes dependencies explicit which is not that bad. Also log do is bad. I tried to explain my position on this here: https://app.slack.com/client/T01UN827FQV/C04C3ASUFEF/thread/C04C3ASUFEF-1680260404.710759

We can leave `let` stuff now and reach consensus on that late.

delegate-app/Main.hs
MangoIV: I think semantically the names are arbitrary and interchangeable. I don't know why the name "channel" would imply that it is implemented as an actual distributed list whereas the Queue is implemented as a single TVar holding a normal list

delegate-app/Main.hs
uhbif19: What is app in it?

delegate-app/Main.hs
uhbif19: Okay.

delegate-app/HydraAuction/Delegate/Tracing.hs
uhbif19: Okay, you are right, that may be actual usecase if we will kill `logMsg`. And we will need ability to make message types non fixed per monad, by embedding one into another. That is also a way to kill `logMsg` - just introduce String message type instance.

delegate-app/Main.hs
uhbif19: Ah, there is some specialized type in config, understand now, you right.

uhbif19: Ah, there is some specific port type in config, understand now, you right.

delegate-app/Main.hs
uhbif19: We use word `Runner` to name monad pattern. You name kind of thread the same term. But it is different meaning for same term - we use `Runner` monad(s) in multiple different situations. That is such term selection is confusing.

src/HydraAuction/Types.hs
uhbif19: > because I like the fish

Yep, I did understood that :D

delegate-app/Main.hs
MangoIV: I'm not 100% confident but I think this is resolved now.

delegate-app/Main.hs
```diff
@@ -152,12 +165,55 @@ runDelegateServer conf = do
...
+    mapConcurrently_
+      (uncurry startWorker)
+      [ (WebsocketThread, wsServer)
+      , (QueueAuctionStageThread, queueAuctionPhases)
+      , (DelegateRunnerThread, delegateRunner)
+      ]
+
+mbQueueAuctionPhases :: TQueue DelegateInput -> TChan DelegateResponse -> DelegateTracerT IO ()
+mbQueueAuctionPhases reqq =
+  liftIO . atomically . (dupTChan >=> getTerms)
```
uhbif19: Please write such things in `do`-notation.

delegate-app/Main.hs
```diff
@@ -152,12 +165,55 @@ runDelegateServer conf = do
...
+    traceQueueEvent :: AuctionTerms -> DelegateTracerT IO ()
+    traceQueueEvent = trace . QueueAuctionPhaseEvent . ReceivedAuctionSet
+
+    getTerms :: TChan DelegateResponse -> STM AuctionTerms
+    getTerms chan =
+      readTChan chan >>= \case
+        AuctionSet terms -> pure terms
+        _ -> getTerms chan
+
+    queueCurrentStage :: AuctionTerms -> IO ()
```
uhbif19: Name is misleading, this is forever-running thread.

delegate-app/Main.hs
```diff
@@ -152,12 +165,55 @@ runDelegateServer conf = do
     pure q
   delegateBroadCastChan <- liftIO . atomically $ newTChan
 
-  _ <-
-    liftIO . forkIO $
-      runServer (show $ dlgt'host conf) (fromIntegral $ dlgt'port conf) $
-        runWithTracer' tracer . delegateServerApp (dlgt'ping conf) delegateInputChan delegateBroadCastChan
-
-  execDelegateRunnerT $ mkRunner (dlgt'tick conf) delegateInputChan delegateBroadCastChan
+  let startWorker :: ThreadSort -> IO () -> IO ()
```
uhbif19: I think this should be standalone function.

Defunctionalization is fun. But I guess this wont cover the difference of queues which will follow input queue will be separated.

delegate-app/Main.hs
```diff
@@ -152,12 +165,55 @@ runDelegateServer conf = do
...
+      [ (WebsocketThread, wsServer)
+      , (QueueAuctionStageThread, queueAuctionPhases)
+      , (DelegateRunnerThread, delegateRunner)
+      ]
+
+mbQueueAuctionPhases :: TQueue DelegateInput -> TChan DelegateResponse -> DelegateTracerT IO ()
+mbQueueAuctionPhases reqq =
+  liftIO . atomically . (dupTChan >=> getTerms)
+    >=> liftIO . (traceQueueEvent *> queueCurrentStage)
+  where
```
uhbif19, line 202: Thanks.

delegate-app/Main.hs
uhbif19: This should be somehow marked by name. `queueCurrentStage` sounds like one-action function.

delegate-app/HydraAuction/Delegate/Server.hs
```diff
@@ -46,6 +49,9 @@ data DelegateServerLog
   | FrontendInput FrontendRequest
   | DelegateOutput DelegateResponse
   | DelegateError DelegateError
+  | StartThread ThreadSort
```
uhbif19: Start/Cancel could be merged I guess.

MangoIV: How would we merge these?

delegate-app/Main.hs
uhbif19: I guess we can separate this to follow-up issue to iterate quickly.

delegate-app/HydraAuction/Delegate/Server.hs
uhbif19: Like `ThreadEvent Kind ThreadSort`

delegate-app/HydraAuction/Delegate/Tracing.hs
uhbif19: Separated this to follow-up issue: https://github.com/mlabs-haskell/hydra-auction/issues/231

delegate-app/Main.hs
MangoIV: well it's a ServerApp, which is a function from `PendingConnection -> IO ()` https://hackage.haskell.org/package/websockets-0.12.7.3/docs/Network-WebSockets.html#t:ServerApp

delegate-app/Main.hs
MangoIV: this is handled by a follow up PR

delegate-app/Main.hs
```diff
@@ -1,62 +1,254 @@
...
+      atomically $ writeTQueue reqq (DelegateEvent $ AuctionStageStarted currentStage)
       currentPosixTime <- POSIXTime <$> currentTimeMilliseconds
       let mSecsLeft = secondsLeftInInterval currentPosixTime (stageToInterval terms currentStage)
       case mSecsLeft of
         Nothing -> pure ()
         Just s -> do
-          threadDelay (fromInteger s * 1000)
-          queueCurrentStage
-mbQueueAuctionPhases _ _ = pure ()
+          threadDelay (fromInteger s * tick)
```
nazrhom: I don't think this is right... The 1000 is meant to convert seconds to milliseconds, so threadDelay waits the correct amount of time

MangoIV: Oh. Thank you. Will fix

delegate-app/Main.hs
nazrhom: I opened #243 to track this

nazrhom:
Looks like all conversations are resolved here. LGTM

nazrhom (approved):

PR #218, opened 2023-03-28
Queue auction stage events in delegate server after a new auction is set

### LINKED ISSUE #157: Trigger automatic actions defined by auction stages in delegate server
This task requires to introduce polling logic placing in queue `HydraEvent`s  which are triggered by:

* `DelegateEvent Start`
* `DelegateEvent (AuctionStageStarted ...)`

Actual implementation is out of scope of this task.

Blocked by #140

nazrhom:


delegate-app/Main.hs
```diff
@@ -5,45 +5,60 @@ import Prelude
...
+mbQueueAuctionPhases _ _ = pure ()
+
+spawnServer :: Chan DelegateInput -> IO ()
+spawnServer _ = forever $ pure ()
 
 main :: IO ()
 main = do
+  events <- newChan
+  -- Write init even in queue
+  writeChan events (DelegateEvent Start)
```
uhbif19: :+1:

delegate-app/Main.hs
```diff
@@ -5,45 +5,60 @@ import Prelude
...
-        "Delegate responses for input: " <> show delegateResponse
-  Nothing -> return ()
-
-delegateTick :: Int
-delegateTick = 1_000
+mbQueueAuctionPhases :: [DelegateResponse] -> Chan DelegateInput -> IO ()
+mbQueueAuctionPhases [AuctionSet terms] events = queueCurrentStage
+  where
+    queueCurrentStage = do
+      currentStage <- currentAuctionStage terms
```
uhbif19, line 44: I think it would read simpler if separated to `queueCurrentStage` and `waitForNextStage` (taking continuation).

delegate-app/Main.hs
```diff
@@ -5,45 +5,60 @@ import Prelude
...
+        Just s -> do
+          void $ async (threadDelay (fromInteger s * 1000) >> queueCurrentStage)
+mbQueueAuctionPhases _ _ = pure ()
+
+spawnServer :: Chan DelegateInput -> IO ()
+spawnServer _ = forever $ pure ()
 
 main :: IO ()
 main = do
+  events <- newChan
```
uhbif19: I think `delegateInputs` is better name, cuz there is `DelegateEvent` type which is just one of cases.

delegate-app/Main.hs
```diff
@@ -5,45 +5,60 @@ import Prelude
...
+      writeChan events (DelegateEvent $ AuctionStageStarted currentStage)
+      currentPosixTime <- POSIXTime <$> currentTimeMilliseconds
+      let mSecsLeft = secondsLeftInInterval currentPosixTime (stageToInterval terms currentStage)
+      case mSecsLeft of
+        Nothing -> pure ()
+        Just s -> do
+          void $ async (threadDelay (fromInteger s * 1000) >> queueCurrentStage)
+mbQueueAuctionPhases _ _ = pure ()
+
+spawnServer :: Chan DelegateInput -> IO ()
```
uhbif19: I guess you could be already start it in your code to show intended way to do that.

delegate-app/Main.hs
```diff
@@ -5,45 +5,60 @@ import Prelude
...
+  where
+    queueCurrentStage = do
+      currentStage <- currentAuctionStage terms
+      writeChan events (DelegateEvent $ AuctionStageStarted currentStage)
+      currentPosixTime <- POSIXTime <$> currentTimeMilliseconds
+      let mSecsLeft = secondsLeftInInterval currentPosixTime (stageToInterval terms currentStage)
+      case mSecsLeft of
+        Nothing -> pure ()
+        Just s -> do
+          void $ async (threadDelay (fromInteger s * 1000) >> queueCurrentStage)
```
uhbif19: I do not understand async here. Should not this just lead to infinite nested thread creation?

I think just the single async with infinite loop should do the job, like in `consumer`:

```
-- This could only happen once per auction
case delegateResponse of
   [AuctionSet terms]) ->  liftIO $ queueCurrentStages delegateInputsQueue
   _ -> pure ()
```

uhbif19 (changes requested):

delegate-app/Main.hs
nazrhom: Eventually `mSecsLeft` will be `Nothing` (when we reach the cleanup phase, as it doesn't have an upper bound).

delegate-app/Main.hs
nazrhom: Sorry, this was a leftover. I had initially stubbed out the server, but I thought it might only create conflicts, since @MangoIV already has a PR up for that

delegate-app/Main.hs
nazrhom: Can you give an example of that?

Remember that `mbQueueAuctionPhases` is happening in the main thread, so we can not really wait for the next stage there.
The idea here is that, only when you get the `AuctionSet` (which should happen only once), you spawn this series of threads, that will eventually stop once the last stage is reached

delegate-app/Main.hs
```diff
@@ -5,45 +5,57 @@ import Prelude
...
-delegateTick :: Int
-delegateTick = 1_000
+import HydraAuction.Delegate.Interface (DelegateResponse (..))
+import HydraAuction.OnChain.Common (secondsLeftInInterval, stageToInterval)
+import HydraAuction.Tx.Common (currentAuctionStage, currentTimeMilliseconds)
+
+consumer :: Chan DelegateInput -> DelegateRunnerT IO ()
+consumer delegateInputs = do
+  -- This will block until an event appears in the queue
+  delegateIn <- liftIO $ readChan delegateInputs
```
MangoIV, line 32: Afaict this does only one step, can we use a queue here and flush the entire queue? This is what I wanted to do, it avoids queuing up too many actions.

delegate-app/Main.hs
```diff
@@ -5,45 +5,57 @@ import Prelude
...
+  delegateResponse <- delegateStep delegateIn
+  liftIO $ mbQueueAuctionPhases delegateResponse delegateInputs
+  -- FIXME: send response to client
```
MangoIV, line 35: This will be implemented in my PR

MangoIV (commented):
This has quite some overlap with my PR 👀 How do we manage that issue?

delegate-app/Main.hs
uhbif19: Yes. What should it break? Thread just will not recurse then and thread gonna stop.

delegate-app/Main.hs
uhbif19: I have written here on that matter. And I do not understand why this task may need more than one additional thread.

https://github.com/mlabs-haskell/hydra-auction/pull/218/files#r1151053101

uhbif19:
@MangoIV 

> This has quite some overlap with my PR eyes How do we manage that issue?

You can just rebase your PR on that PR and make your PR stacked on that one, no?

delegate-app/Main.hs
uhbif19: I do not understand what is wrong with one step. Like is not the purpose of queue exactly to process inputs one by one?

MangoIV: conceptually, I think we currently run some action after each frontendrequest that can potentially need some amount of time, while that happens, the queue piles up, I'm just wondering whether, if we don't spawn specific handler threads, we're gonna have more `FrontendRequest`s per time than DelegateSteps per time.

delegate-app/Main.hs
```diff
@@ -5,45 +5,57 @@ import Prelude
...
+  liftIO $
+    putStrLn $
+      "Delegate responses for input: " <> show delegateResponse
+
+mbQueueAuctionPhases :: [DelegateResponse] -> Chan DelegateInput -> IO ()
```
MangoIV, line 40: Why did you decide for Chan instead of TChan?

nazrhom: No real reason, we don't really need complex atomic blocks (yet). Feel free to upgrade to a TChan if needed

delegate-app/Main.hs
nazrhom: I think we wont run in this issue with the demo, but it might definitely be a problem for a real world application.
I propose we stick to a single reader/multiple writers for now.. it might turn out that relaxing to multiple readers is trivial, and we can do that at the end (or it might turn out to be hard, and we can ignore it 😄 )

uhbif19: > if we don't spawn specific handler threads

I do not think that will help, cuz Delegate server does not do anything heavy, it just creates requests to Hydra node. So that is essentially non-blocking and non-IO-heavy task, so I do not understand why it would gain performance by adding thread.

> but it might definitely be a problem for a real world application

I do not think where should be such situation even in production. Ideally we have ~1000 ticks a second now. So if one bid requires like 10 inputs which seems like very pessimistic count to me, we can support 100 clients bidding every second. This seems like a lot to me.

Possibly we might get much less ticks it real work takes time.

I think that @MangoIV point is important in that we need to:

* Document this potential queue overloading issue in our limitations list
* Add some logging or other way to debug such situation

delegate-app/Main.hs
nazrhom: I was answering `Should not this just lead to infinite nested thread creation?`. But I understand your point about not spawning many nested threads, I refactored this so that it is all happening in a single async action.

delegate-app/Main.hs
MangoIV: I think the way I've done in in my PR is perhaps a solution? Can you look at that?

MangoIV:
> You can just rebase your PR on that PR and make your PR stacked on that one, no?

I have got more commits and more significant changes to the Main logic. Wouldn't it be better to do it vice Versa?

uhbif19:
@MangoIV 

> I have got more commits and more significant changes to the Main logic. Wouldn't it be better to do it vice Versa?

This PR is nearly finished. And actually I missed that @nazrhom already answered my concerns, and probably gonna merge it now.

Your PR marked as draft so I did not checked last version.

uhbif19 (approved):

Issue #180, opened 2023-03-09
Support OS X in our demo


uhbif19:
Currently we use `docker-compose`. On OS X socket files which `cardano-node` use does not work in it.

So currently we suggest to use VM to start project on OS X. If there will be demand for OS X support we could think of more convenient solution.

**UPD:**

Current plan is try to run REPL inside containers, as hydra does. If it works: provide `Makefile` shortcut and add to README.

nazrhom:
This is partially solved by the PR above. There is now an image for the CLI as well that can be built through nix. This image is not included in the compose file, but can be run on its own by a mac os user to be able to speak to the cardano-node

PR #219, opened 2023-03-28
Add derivations to build docker images for cli and delegate server.

nazrhom:
This allows us to make some progress on #180 as osx users can now run the cli and delegate server through docker (thus fixing the socket issue).

There is still the caveat that building this derivation requires access to a linux machine (i.e. remote nix builder)


MangoIV:
where did my changes wrt cross go? are they on another branch that hasn't been merged yet?

MangoIV:
also, do you want me to change it so it's only buildable on linux liked you asked internally?

nazrhom:
@MangoIV I created another branch, the changes you did for cross-compilation should still be in the other one

nazrhom:
> also, do you want me to change it so it's only buildable on linux liked you asked internally?

Yeah, please! I think it makes sense to only add it for the platforms on which it would work

scripts/spin-up-new-devnet.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh -e
+#!/bin/bash -e
```
MangoIV:
```suggestion
#!/usr/bin/env bash -e
```

scripts/stop-demo.sh
```diff
@@ -2,3 +2,4 @@
 
 docker-compose down
 docker-compose --profile hydra-node down
+docker-compose --profile delegate-server down
```
uhbif19, line 3: Should not it be the other way around? Like delegate server depends on hydra-node, so should be killed first.

And I still do not know what is wrong with `depends_on`, which seems to do the same thing automatically.   I remember there was some problem with it... Like it does not ensure container started by health-check.

docker-compose.yaml
```diff
@@ -120,6 +120,51 @@ services:
         ipv4_address: 172.16.238.30
     restart: always
 
+  devnet-server-1:
+    image: hydra-auction-delegate:latest
```
uhbif19, line 124: So, I am missing:

* Where does building take place?
* Where do it stores images? Locally I presume?

Does our CI automatically call whose derivations building images?

uhbif19: Could you please explain how to call building locally.

scripts/stop-demo.sh
nazrhom: `depends_on` is a bit tricky, as it does not care that whatever is running inside the container is "ready". I.e. we have to manually wait for the socket to be present from cardano-node, before running anything else...

uhbif19: Seems like now is `condition: service_healthy` for that. While I am not sure it actually works well.

Anyway, our current code does not seem to have any waits, so I works the same as plain depends_on.

uhbif19 (approved):

README.md
```diff
@@ -38,9 +38,16 @@ Will build the CLI application and link it to `./result/bin/hydra-auction`.
...
+which will start a single cardano-node,
+3 hydra-nodes for the delgates,
+and 3 instances of the delegate server.
+
+Before starting the demo we will need to build the delegate server docker images.
+This is done through nix and can be done by `make build-docker`.
+This command will build the image and load it as `hydra-auction-delegate:latest`.
 
 We recommend using the `spin-up-new-devnet.sh` script
+to run the demo and start the components,
```
uhbif19, line 50: I think "the demo" is the whole process, including what we do in frontend cli manually.

PR #225, opened 2023-03-28
Add Fixture.Actors for Hydra node and Fixture.partyFor and something else

uhbif19:


hydra-auction.cabal
```diff
@@ -22,6 +22,7 @@ data-files:
   data/credentials/grace.vk
   data/credentials/hans.sk
   data/credentials/hans.vk
+  data/hydra-keys/*
```
nazrhom: I don't think you can use the `*` here. iirc you need to pass the fully qualified paths

src-utils/HydraAuctionUtils/Fixture.hs
```diff
@@ -42,39 +48,59 @@ data Actor
...
+  bs <-
+    readDataFile $
+      "hydra-keys" </> actorName actor <.> "vk"
+  let res =
+        first TextEnvelopeAesonDecodeError (Aeson.eitherDecodeStrict bs)
+          >>= deserialiseFromTextEnvelope (AsVerificationKey AsHydraKey)
+  case res of
+    Left err ->
+      fail $ "cannot decode text envelope from '" <> show bs <> "', error: " <> show err
+    Right vk -> return $ Party vk
```
nazrhom, line 95: nit: I think we generally use `pure` over `return`

src-utils/HydraAuctionUtils/Fixture.hs
```diff
@@ -42,39 +48,59 @@ data Actor
...
 
+-- These actors are supposed to be used as Hydra node admins
+hydraNodeActors :: [Actor]
+hydraNodeActors = [Rupert, Oscar, Patricia]
+
 -- | Get the "well-known" keys for given actor.
 keysFor :: Actor -> IO (VerificationKey PaymentKey, SigningKey PaymentKey)
 keysFor actor = do
-  bs <- readConfigFile ("credentials" </> actorName actor <.> "sk")
+  bs <- readDataFile $ "credentials" </> actorName actor <.> ".sk"
```
nazrhom: I think `<.>` should already include the `.`

src-utils/HydraAuctionUtils/Fixture.hs
uhbif19: I always not follow that rule, and even disabled that in Hlint.

I understand what adding it to language in first place was wrong, but do not see what is wrong with its usage, cuz it adds more information on type.

But we can enforce pure usage if somebody has really bad feelings about return, but in that case we need to fix it everythere.

hydra-auction.cabal
uhbif19: I checked, docs say you can: https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-data-files

Also it is covered by test.

nazrhom (approved):

nazrhom:
This looks good to me, I am not sure why the tests are failing though... 
I see

```
FAIL
      Exception: SubmitTxValidationError (TxValidationErrorInMode (ShelleyTxValidationError ShelleyBasedEraBabbage (ApplyTxError [UtxowFailure (UtxoFailure (FromAlonzoUtxoFail (OutsideValidityIntervalUTxO (ValidityInterval {invalidBefore = SJust (SlotNo 80), invalidHereafter = SJust (SlotNo 110)}) (SlotNo 110))))])) BabbageEraInCardanoMode)
```

in the logs.. are our tests flaky? I don't see anything in this PR that should cause that

PR #226, opened 2023-03-29
Add monad interfaces instances for Hydra and fix Hydra.Interface

### LINKED ISSUE #125: Add support for Hydra in off-chain util functions

uhbif19:
Closes #125 

Blocked by #225


src/HydraAuction/Hydra/Interface.hs
```diff
@@ -1,3 +1,74 @@
...
+  | TxSeenKind
+  | CommittedKind
+  | HeadIsInitializingKind
+  | HeadIsOpenKind
+  | HeadIsClosedKind
+  | ReadyToFanoutKind
+  | HeadIsFinalizedKind
+  deriving stock (Eq)
+
+getHydraEventKind :: HydraEvent -> HydraEventKind
```
nazrhom, line 71: I think we can use makePrisms for `HydraEvent` to achieve something similar. We can also remove the reification of `AwaitedHydraEvent` and make it `type AwaitedHydraEvent = HydraEvent -> Bool` so the generated functions can be used to check which constructor is chosen

uhbif19: > I think we can use makePrisms for HydraEvent to achieve something similar.

Thanks, that seems like exactly the question I was talking about in haskell-learn channel.

I would not change that now, cuz I am bad at prisms, but you could refactor that later is you want to.

> make it type AwaitedHydraEvent = HydraEvent -> Bool so the generated functions can be used to check which constructor is chosen

Functions do not have Eq and are more obscure to understand. Also ADT AwaitedHydraEvent is able have some other implementation than filtering, in theory.

src/HydraAuction/Hydra/Interface.hs
```diff
@@ -1,3 +1,74 @@
...
+data HydraCommand
+  = Init
+  | GetUTxO
+  | NewTx Tx
+  | Commit UTxO.UTxO
+  | Close
+  | Fanout
+  deriving stock (Show)
+
+commandConstructorName :: HydraCommand -> Text
```
MangoIV, line 34: This can be done cleanly with Generics to avoid repetitions

src/HydraAuction/Hydra/Runner.hs
```diff
@@ -0,0 +1,152 @@
...
+import HydraAuction.Hydra.Monad (
+  AwaitedHydraEvent (..),
+  MonadHydra (..),
+ )
+import HydraAuctionUtils.Monads (
+  MonadNetworkId (askNetworkId),
+  MonadTrace (..),
+ )
+
+{- HLINT ignore "Use newtype instead of data" -}
```
MangoIV, line 63: This will trigger an unknown Pragma warning, do we have that warning enabled globally? I think it should be enabled per module.

src/HydraAuction/Hydra/Interface.hs
MangoIV: i think this can be done easier by having an NS that is tagged with a constructor name. This would be possible to be created generically. 

E.g. you could collect the constructor names of the type from its Rep in a list then have a GADT that witnesses that list.  The GADT could looks similar to NS from generics-sop

MangoIV: Perhaps using lens is the better solution though.

src/HydraAuction/Hydra/Runner.hs
uhbif19: This is not a pragma, this is comment. For exactly that reason.

MangoIV: OH, sorry then. the HLS action for this always adds a pragma and another pragma to ignore the hlint pragma, so that's why I'm used to seeing that.

src/HydraAuction/Hydra/Interface.hs
uhbif19: It could, but seems like quite overkill. Also we need ability to change encodings in case of Hydra API changes.

src/HydraAuction/Hydra/Monad.hs
```diff
@@ -0,0 +1,72 @@
...
+
+waitForHydraEvent :: MonadHydra m => AwaitedHydraEvent -> m HydraEvent
+waitForHydraEvent = waitForHydraEvent' defaultTimeout
+
+sendCommandAndWaitFor :: MonadHydra m => AwaitedHydraEvent -> HydraCommand -> m HydraEvent
+sendCommandAndWaitFor awaitedSpec command = do
+  sendCommand command
+  waitForHydraEvent awaitedSpec
+
+instance (Monad m, MonadHydra m) => MonadSubmitTx m where
```
MangoIV: `Monad m` should be subsumed by `MonadHydra`, but that's not so important

src/HydraAuction/Hydra/Monad.hs
```diff
@@ -0,0 +1,72 @@
...
+  submitTx :: Tx -> m ()
+  submitTx = sendCommand . NewTx
+  awaitTx :: Tx -> m ()
+  awaitTx = void . waitForHydraEvent . SpecificEvent . TxSeen
+
+instance {-# OVERLAPPABLE #-} (Monad m, MonadHydra m) => MonadQueryUtxo m where
```
MangoIV, line 78: This is probably not good practice, there's two possibilities to solve this in a better way: 
- use a `newtype` and `deriving via` 
- make a `default` declaration and use it with `deriving anyclass`

src/HydraAuction/Hydra/Runner.hs
```diff
@@ -0,0 +1,152 @@
...
+  HydraRunner a ->
+  IO a
+executeHydraRunner context runner =
+  runReaderT (unHydraRunner runner) context
+
+executeHydraRunnerInTest :: HydraClient -> HydraRunner a -> IO a
+executeHydraRunnerInTest node = executeHydraRunner context
+  where
+    tracer = contramap show stdoutTracer
+    context = MkHydraExecutionContext {node = node, tracer = tracer}
```
MangoIV, line 174:
```suggestion
    context = MkHydraExecutionContext node tracer
```

src/HydraAuction/Hydra/Runner.hs
```diff
@@ -0,0 +1,152 @@
...
+instance MonadTrace HydraRunner where
+  type TracerMessage HydraRunner = HydraRunnerLog
+  stringToMessage = HydraRunnerStringMessage
+  traceMessage message = do
+    MkHydraExecutionContext {tracer} <- ask
+    liftIO $ traceWith tracer message
+
+matchingHydraEvent :: Value -> Maybe HydraEvent
+matchingHydraEvent value =
+  case value ^? key "tag" of
```
MangoIV, line 132: I think with `Maybe` this will be kinda bad to debug, I think it would be nice to have a real Parser or some more informative error type instead

src/HydraAuction/Hydra/Monad.hs
```diff
@@ -0,0 +1,72 @@
...
+    response <-
+      sendCommandAndWaitFor (SpecificKind GetUTxOResponseKind) GetUTxO
+    let utxo = case response of
+          GetUTxOResponse utxo' -> utxo'
+          _ -> error "Impossible happened: incorrect response type awaited"
+    return $ UTxO.fromPairs $ filter predicate $ UTxO.pairs utxo
+    where
+      predicate (txIn, TxOut (AddressInEra _ txOutAddress) _ _ _) = case query of
+        -- FIXUP: use safe coerce
+        ByAddress address -> txOutAddress == unsafeCoerce address
```
MangoIV: why is this `unsafeCoerce`? I feel like this should at least be annotated with specific types such that it won't silently blow up.

MangoIV: This cannot be safely coerced, ever.

MangoIV:
```suggestion
        ByAddress address ->
          case txOutAddress of 
            addr@(ShelleyAddress _ _ _) -> 
              address == addr 
            ByronAddress _ -> error "didn't expect Byron address" 
```

src/HydraAuction/Hydra/Monad.hs
```diff
@@ -0,0 +1,72 @@
...
+  sendCommandAndWaitFor,
+) where
+
+-- Prelude imports
+import Prelude
+
+-- Haskell imports
+import Control.Monad (void)
+import GHC.Natural (Natural)
+import Unsafe.Coerce (unsafeCoerce)
```
MangoIV:
```suggestion
```

src/HydraAuction/Hydra/Monad.hs
```diff
@@ -0,0 +1,72 @@
...
+
+-- Cardano imports
+import Cardano.Api (AddressInEra (..))
+import Cardano.Api.UTxO qualified as UTxO
+
+-- Hydra imports
+import Hydra.Cardano.Api (Tx)
```
MangoIV:
```suggestion
import Hydra.Cardano.Api (Tx, Address (ByronAddress, ShelleyAddress))
```

src/HydraAuction/Hydra/Monad.hs
MangoIV: This can also be some real error with something like `filterM`

MangoIV (commented):
Fix `unsafeCoerce`

src/HydraAuction/Hydra/Monad.hs
MangoIV: it's not as unsafe as I thought, because at least it's not a segfault, but I still think we should not do it. https://play.haskell.org/saved/UElf70yi

src/HydraAuction/Hydra/Runner.hs
uhbif19: This is not a parser and it does not have errors. This is matcher to await events.

src/HydraAuction/Hydra/Monad.hs
uhbif19: Thanks, will merge.

src/HydraAuction/Hydra/Monad.hs
uhbif19: This does not work, I do not know why.

> Could not deduce (Monad m)
>        arising from the superclasses of an instance declaration

uhbif19: Lol, seems to be known issue, and examples are having the same Monad superclass too :D

https://gitlab.haskell.org/ghc/ghc/-/issues/14836

uhbif19: Okay, I am wrong, UndecidableSuperClasses does not help. Which is quite understandable, cuz there is nothing recursive here.

I will just add FIXME.

src/HydraAuction/Hydra/Monad.hs
uhbif19: I do not see which problems could OVERLAPPABLE introduce. Especially given that we do all instances in same package. And MLabs styleguide does not mention overlapping.  And we already use bunch of Orphan instances.

It builds now without OVELAPPABLE now though. I am now sure why, probably we do not import both instances in same file.

src/HydraAuction/Hydra/Runner.hs
uhbif19: Named fields are better, cuz they provide explanation on arguments.

src/HydraAuction/Hydra/Monad.hs
uhbif19: Because I was lazy, will apply your patch, thanks.

src/HydraAuction/Hydra/Monad.hs
MangoIV: There's a couple of reasons why this is bad: 
- behavior may change by omitting an instance declaration 
- especially in combination with orphan instances this is incredibly dangerous
- I think it's generally best practice to not reach for OVERLAPPING and OVERLAPPABLE without good reasons, I don't see a good reason here because we could just have a `default` impl or a `newtype` wrapper, which are very nice solutions 
- also, that it works at the moment is because you don't use the instance. overlapping and overlappable errors are discovered at call sites.

uhbif19: > also, that it works at the moment is because you don't use the instance

I think I checked this with instances.

src/HydraAuction/Hydra/Monad.hs
MangoIV:
```haskell

Main.hs:5:10: warning: [GHC-36038] [-Wloopy-superclass-solve]
    I am solving the constraint ‘Monad m’,
      arising from the superclasses of an instance declaration,
    in a way that might turn out to loop at runtime.
    Starting from GHC 9.10, this warning will turn into an error.
    See the user manual, § Undecidable instances and loopy superclasses.
    Suggested fix:
      Add the constraint ‘Monad m’ to the instance context,
      even though it seems logically implied by other constraints in the context.
```

MangoIV: this is the reason.

MangoIV: https://play.haskell.org/saved/WSLu5ueh

MangoIV:
```mermaid
stateDiagram-v2
    Monad --> MonadA
    Monad --> MonadB 
```

uhbif19: I forgot we have UndecidableInstances enabled globaly.

MangoIV:
```mermaid
stateDiagram-v2
    Monad --> MonadA
    Monad --> MonadB 
    MonadA --> MonadB
```

MangoIV: the above is the class relations and the below is the instance relation. Perhaps the reason is that it won't know whether to pick the instance that is implied by MonadB or the one that is implied by MonadA?

src/HydraAuction/Hydra/Monad.hs
uhbif19: Not sure how to deal with `newtype` for case of `* -> *` kinded types here. But you probably right that it is cleaner.

Please separate that to follow-up tech-debt issue.

src/HydraAuction/Hydra/Runner.hs
MangoIV: then this is really strange to me.

src/HydraAuction/Hydra/Monad.hs
MangoIV: `newtype Wrap f a = Wrap (f a)`

MangoIV: however, you can also do a `default` decl

MangoIV (approved):

src/HydraAuction/Hydra/Runner.hs
nazrhom: I think this makes sense as a `Maybe`. We don't really know what to do with the error here. We want to run this over every event, to build a `HydraEvent` and then run a predicate on it. Using `Maybe` means we ignore every JSON value coming in that does not match what we expect.

src/HydraAuction/Hydra/Monad.hs
```diff
@@ -0,0 +1,93 @@
...
+waitForHydraEvent :: MonadHydra m => AwaitedHydraEvent -> m HydraEvent
+waitForHydraEvent = waitForHydraEvent' defaultTimeout
+
+sendCommandAndWaitFor :: MonadHydra m => AwaitedHydraEvent -> HydraCommand -> m HydraEvent
+sendCommandAndWaitFor awaitedSpec command = do
+  sendCommand command
+  waitForHydraEvent awaitedSpec
+
+-- FIXME: this does not work without `Monad m`, do not know why
+instance {-# OVERLAPPABLE #-} (Monad m, MonadHydra m) => MonadSubmitTx m where
```
nazrhom, line 66: I read the discussion in the other comment, I agree we can defer this to a follow-up PR. Maybe we could just add a comment here explaining why we are using overlappable

nazrhom (approved):
Looks good to me! Just added a note about explaining why we have the OVERLAPPABLE pragma

src/HydraAuction/Hydra/Monad.hs
uhbif19: Follow-up #237

PR #229, opened 2023-04-01
Introduce MonadBlockchainParams and separate part of `Tx.Commons`

uhbif19:
Follow up for #125 
Part of #178 
Blocked by #226

### DIFF
--- a/src-utils/HydraAuctionUtils/Tx/AutoCreateTx.hs
+++ b/src-utils/HydraAuctionUtils/Tx/AutoCreateTx.hs
@@ -1,0 +1,201 @@
+{-# LANGUAGE RecordWildCards #-}
+
+module HydraAuctionUtils.Tx.AutoCreateTx (
+  AutoCreateParams (..),
+  autoCreateTx,
+  autoSubmitAndAwaitTx,
+) where
+
+-- Prelude imports
+import Prelude
+
+-- Haskell imports
+import Data.Foldable (Foldable (toList))
+
+-- Cardano imports
+import Cardano.Api.UTxO qualified as UTxO
+
+-- Plutus imports
+import Plutus.V2.Ledger.Api (POSIXTime)
+
+-- Hydra imports
+import Hydra.Cardano.Api (
+  Address,
+  BuildTx,
+  BuildTxWith,
+  CtxTx,
+  KeyWitness,
+  PaymentKey,
+  ShelleyAddr,
+  SigningKey,
+  Tx,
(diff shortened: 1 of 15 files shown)


src-utils/HydraAuctionUtils/Tx/AutoCreateTx.hs
```diff
@@ -0,0 +1,201 @@
...
+  pattern TxReturnCollateralNone,
+  pattern TxScriptValidityNone,
+  pattern TxTotalCollateralNone,
+  pattern TxUpdateProposalNone,
+  pattern TxValidityLowerBound,
+  pattern TxValidityNoLowerBound,
+  pattern TxValidityNoUpperBound,
+  pattern TxValidityUpperBound,
+  pattern TxWithdrawalsNone,
+  pattern WitnessPaymentKey,
```
MangoIV, line 69: are these exported together with some datatype? IOG never misses to disappoint, having this in every module is horrible. Imagine if we didn't have HLS, this would be unusable.

MangoIV (approved):

src-utils/HydraAuctionUtils/Tx/AutoCreateTx.hs
uhbif19: > having this in every module is horrible

That is one of the reasons we have this utility handling that stuff in one place (except the `commitTx` hacks).

Also I written most of this utility before having HLS working :D

uhbif19: In theory one should use `plutus-apps` and not use this API directly.

MangoIV: without HLS this is just plain horrifying.

uhbif19: Also GHC has problems with `pattern` importing errors: https://gitlab.haskell.org/ghc/ghc/-/issues/18466

That affects HLS too.

src-utils/HydraAuctionUtils/Tx/AutoCreateTx.hs
```diff
@@ -0,0 +1,201 @@
+{-# LANGUAGE RecordWildCards #-}
```
uhbif19, line 1: @mlabs-haskell/hydra-auction 

That is moved file!

Most of stuff are the same. All logical changes you can see in separate `commit` before moving

PR #234, opened 2023-04-03
Separate `delegateStep` into two parts for per-client response support

uhbif19:
Part of #175 

Blocked by #226

### DIFF
--- a/delegate-app/Main.hs
+++ b/delegate-app/Main.hs
@@ -17,32 +17,32 @@ import Plutus.V2.Ledger.Api (POSIXTime (..))
 
 import HydraAuction.Delegate (
   DelegateEvent (..),
-  DelegateInput (DelegateEvent),
   DelegateRunnerT,
-  delegateStep,
+  delegateEventStep,
   execDelegateRunnerT,
  )
 import HydraAuction.Delegate.Interface (DelegateResponse (..))
 import HydraAuction.OnChain.Common (secondsLeftInInterval, stageToInterval)
 import HydraAuction.Tx.Common (currentAuctionStage, currentTimeMilliseconds)
 
-consumer :: Chan DelegateInput -> DelegateRunnerT IO ()
-consumer delegateInputs = do
+consumer :: Chan DelegateEvent -> DelegateRunnerT IO ()
+consumer delegateEvents = do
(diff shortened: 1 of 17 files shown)


src/HydraAuction/Delegate.hs
```diff
@@ -57,37 +51,51 @@ execDelegateRunnerT (MkDelegateRunner action) = evalStateT action initialState
...
         -- FIXME: place new bid
         -- FIXME: return closing transaction
-        return [ClosingTxTemplate]
+        return [(PerClient clientId, ClosingTxTemplate)]
+
+delegateEventStep ::
+  forall m.
+  Monad m =>
+  DelegateEvent ->
+  DelegateRunnerT m [DelegateResponse]
```
nazrhom, line 88: Maybe we could make this `DelegateRunnerT m ()`. We only need the delegate response for the front end clients.

delegate-app/Main.hs
```diff
@@ -17,32 +17,32 @@ import Plutus.V2.Ledger.Api (POSIXTime (..))
...
-consumer delegateInputs = do
+consumer :: Chan DelegateEvent -> DelegateRunnerT IO ()
+consumer delegateEvents = do
   -- This will block until an event appears in the queue
-  delegateIn <- liftIO $ readChan delegateInputs
-  delegateResponse <- delegateStep delegateIn
-  liftIO $ mbQueueAuctionPhases delegateResponse delegateInputs
+  -- FIXME: support frontend requests too
+  delegateEvent <- liftIO $ readChan delegateEvents
+  delegateResponse <- delegateEventStep delegateEvent
```
nazrhom, line 33: Maybe we could remove the return type from `delegateEventStep`, and move `mbQueueAuctionPhases` to the frontend request handler then. We don't really need to respond to anything when handling delegate events.

uhbif19: I only changed Main to pass CI, @MangoIV PR #217 gonna rewrite Main.

> We don't really need to respond to anything when handling delegate events.

That is not correct, we need to broadcast things from `delegateEventStep`. Like New bid or Auction state change, for example.

nazrhom (approved):

PR #242, opened 2023-04-03
Updated  "real network interface ... for delegate server" with per-client messages support

uhbif19:
Based on @MangoIV  PR #217 

Changes:

* Merged staging and #234 
* Support for per-client and broadcast messages
* Some style and naming changes (sorry)

Docstrings may be out of sync now.


delegate-app/Main.hs
```diff
@@ -1,62 +1,336 @@
...
+                writeTQueue frontendRequestQueue (clientId, request)
+      sendFromChannel ::
+        Connection ->
+        ClientId ->
+        TChan (ClientResponseScope, DelegateResponse) ->
+        IO ()
+      sendFromChannel connection clientId broadcastCopy = do
+        (scope, message) <- atomically (readTChan broadcastCopy)
+        when (clientIsInScope clientId scope) $
+          sendToClient connection message
```
MangoIV, line 158: I think it was good documentation to have it return an IO void to signal that it's running infinitely, now with `forever` in the block, you will have to read the whole `do` block again, wasn't that what you were trying to avoid?

delegate-app/Main.hs
```diff
@@ -1,62 +1,336 @@
...
+{- | create a delegate runner that will listen for incoming events
+     and output outgoing events accordingly
+-}
+runDelegateLogicSteps ::
+  forall void.
+  -- | the time in milliseconds that the runner sleeps between acts
+  Int ->
+  -- | the queue of incoming messages
+  TQueue DelegateEvent ->
+  TQueue (ClientId, FrontendRequest) ->
```
MangoIV, line 169: why do we have two Queues here, I don't understand how we need more than one queue, why are frontendrequests not an "event" anymore?

delegate-app/Main.hs
```diff
@@ -1,62 +1,336 @@
...
+    mapConcurrently_
+      (runWithTracer' tracer . workerAction)
+      [ WebsocketThread
+      , QueueAuctionStageThread
+      , DelegateRunnerThread
+      ]
+  where
+    freshClientIdGenerator clientCounter = do
+      v <- takeMVar clientCounter
+      putMVar clientCounter (v + 1)
```
MangoIV, line 256: why do we not return the fresh one but the old one? that sounds dangerous if we don't wrap the `MVar`

delegate-app/Main.hs
```diff
@@ -1,62 +1,336 @@
...
+      -- that sending a ping every 30 seconds is a good idea.
+      withPingThread connection pingSecs (pure ()) $
+        race_
+          (forever $ receiveAndPutInQueue tracer connection clientId)
+          ( forever $
+              sendFromChannel connection clientId toClientsChannelCopy
+          )
+    where
+      sendToClient :: forall a. ToJSON a => Connection -> a -> IO ()
+      sendToClient connection = sendTextData connection . encode
```
MangoIV, line 132: this name has less information than before, `sendTextData` also sends to client, I don't know how this is better than `encodeSend` which at least expresses all it does.

MangoIV (changes requested):

delegate-app/Main.hs
uhbif19: What is the difference? This changes atomicity somehow?

delegate-app/Main.hs
uhbif19: With `DelegateEvent` `devegateStep` typing seems to be wrong for broadcasting. You can see why in #234

delegate-app/HydraAuction/Delegate/Server.hs
```diff
@@ -0,0 +1,131 @@
...
+type DelegateTracerT = TracerT DelegateServerLog
+
+-- | like @ServerApp@ but can be used with a transformer
+type ServerAppT m = PendingConnection -> m ()
+
+-- | a newtype for deriving Pretty via Show
+newtype ViaShow a = ViaShow {unViaShow :: a}
+  deriving stock (Eq, Ord, Generic)
+
+instance Show a => Pretty (ViaShow a) where
```
nazrhom, line 126: I guess the issue is if we say `Show a => Pretty a` then we make instance resolution non-deterministic... but why can't GHC pick a `Pretty` instance if it has one in scope, and only if it can't find one then follow the `Show` branch?

MangoIV: Because you don’t want that. This is the motivation for deriving via. You don’t want to have catch all instances because of the open world assumption of Haskell classes.

delegate-app/Main.hs
```diff
@@ -1,62 +1,336 @@
...
+mbQueueAuctionPhases ::
+  TQueue DelegateEvent ->
+  TChan (ClientResponseScope, DelegateResponse) ->
+  DelegateTracerT IO ()
+mbQueueAuctionPhases delegateEvents toClientsChannel = do
+  terms <-
+    liftIO . atomically $
+      awaitForAuctionSetTerms =<< dupTChan toClientsChannel
+  traceQueueEvent terms
+  liftIO $ forever $ queueCurrentStageAndWaitForNext terms
```
nazrhom: Why is this `forever`? This eventually reduces to `forever $ pure ()`.

src/HydraAuction/Delegate.hs
```diff
@@ -57,37 +54,56 @@ execDelegateRunnerT (MkDelegateRunner action) = evalStateT action initialState
...
-  -- FIXME: abort Hydra node
-  DelegateEvent (AuctionStageStarted VoucherExpiredStage) -> return []
-  DelegateEvent (AuctionStageStarted _) -> return []
-  -- FIXME: commit empty transaction
-  DelegateEvent (HydraEvent (Committed _)) -> return []
-  -- FIXME: fanout Hydra node
-  DelegateEvent (HydraEvent ReadyToFanout) -> return []
-  DelegateEvent (HydraEvent _) -> return []
+  (ClientId, FrontendRequest) ->
+  DelegateRunnerT m [(ClientResponseScope, DelegateResponse)]
```
nazrhom, line 72: Why do we have a list in the return type here? I think the current interface (Broadcast/Perclient) lets us do everything we need.

uhbif19: In case of returning multiple responses. Like one for broadcast and one per-request. It is not used now, but maybe there is some case where this is required. Probably we do not need it, but it does not seem to hurt now, no?

MangoIV: I mean it’s a bit a distinction between a list of pairs and a pair of lists. (Besides perhaps performance characteristics)

delegate-app/Main.hs
uhbif19: You right, I missed that.

delegate-app/Main.hs
uhbif19: All local functions send something to queue or to client. So while I was reading this code I tried to understand which part send where.

And encoding to text or to bytecode is implementation detail which is only important to client, and does not matter for design of this module.

delegate-app/HydraAuction/Delegate/Server.hs
uhbif19: > but why can't GHC pick a Pretty instance if it has one in scope, and only if it can't find one then follow the Show branch?

Is not that what the OVERLAPPABLE does?

> newtype ViaShow

It is strange to me, that Pretty does not already have that implemented. Probably cuz `derviving via` is a recent thing.

MangoIV: The good thing is that the whole point of deriving via is that we do not have to care for whether it's implemented already, we can adhoc deriveable instances by just defining a newtype wrapper

src/HydraAuction/Delegate.hs
uhbif19: Pair of lists does not seem to work here, cuz we need to process in pairs here.

Also we may need change this pair into ADT later, cuz `delegateXStep` may do other kinds of responses, like adding DelegateEvent.

delegate-app/Main.hs
uhbif19: > think it was good documentation to have it return an IO void to signal that it's running infinitely

Yes.

> you will have to read the whole do block again

But how one should get to `send/receive` functions signature other way then reading do-block? They are local functions.

delegate-app/Main.hs
nazrhom: Im curious about this as well, what scenario are you thinking of @MangoIV?

src/HydraAuction/Delegate.hs
nazrhom: @uhbif19 I think you are right. As a matter of fact, in your next PR I see a good case for returning multiple responses.

When a client places a `newBid` we want to: confirm the bid is placed and give the user the ClosingTxTemplate, but also broadcast an update of the delegate state.

delegate-app/Main.hs
MangoIV: it's not so important, but if we take out the MVar and expect to get the same next value in the counter, then that's an incorrect assumption because it will be the next next value.

uhbif19: > and expect to get the same next value in the counter

Why would somebody except that?

MangoIV (approved):

PR #241, opened 2023-04-03
Hydra e2e test

uhbif19:
Closes #124 
Closes #144 
Closes #79 

Blocked by #240 

CI broken by #249

PR #259, opened 2023-04-10
Some delegate  refactoring

uhbif19:
Main change - I removed `DelegateRunnerT`, cuz it is just StateT now. First I done renaming and separated it to module, but then I understood that it makes no sense anymore.

Other changes may be seem in commit messages (after filtering merged #241 commits, lol)

Part of #158 

Blocked by #247 (stacked on it)
Blocked by #241  (merged here)


src/HydraAuction/Delegate.hs
```diff
@@ -40,25 +36,6 @@ data DelegateEvent
   | AuctionStageStarted AuctionStage
   | HydraEvent HydraEvent
 
-newtype DelegateRunnerT m x = MkDelegateRunner
```
nazrhom, line 43: Maybe we could keep around a type synonym (like: `type DelegateRunnerT m a = StateT DelegateState m a`, just in case we decide to add other stuff in the future

nazrhom (approved):

src/HydraAuction/Delegate.hs
uhbif19: Good thought, but I oppose:

1. MLabs styleguide discourages `type`s usage
2. Change gonna be simple either way, this pattern should not go in other places

PR #260, opened 2023-04-10
Implement bidder deposit.

### LINKED ISSUE #246: Implement bidder deposit logic
Logic is described in on chain spec.

Tests are part of scope of this issue.
If that will introduce not problems it is better to also introduce unit tests using PSM.
If not - just continue with our `EndToEnd.Leger` tests.

Todo:

- [x] On chain, tx construction and tests

### LINKED ISSUE #257: Add bidder deposit creation to Frontend CLI
Blocked by #246

nazrhom:
- [x] Losing Bidder
- [x] Seller Claims
- [x] Winning Bidder
- [x] Cleanup


src/HydraAuction/Tx/Deposit.hs
```diff
@@ -0,0 +1,138 @@
...
+  let mp = policy terms
+      voucerCS = VoucherCS $ scriptCurrencySymbol mp
+      bidDepositDatum = BidDepositDatum (toPlutusKeyHash $ verificationKeyHash bidderVk) voucerCS
+      bidDepositTxOut =
+        TxOut
+          (ShelleyAddressInEra depositAddress)
+          depositValue
+          (mkInlineDatum bidDepositDatum)
+          ReferenceScriptNone
+      -- FIXME: take as input?
```
nazrhom: Should we use a fixed deposit value, or should we allow bidders to pick their own?

uhbif19: They should pick.

uhbif19: And the logic for selection of bidders will be completely off-chain.

doc/on_chain_spec.md
```diff
@@ -526,6 +526,11 @@ inform prospective bidders in the auction announcement
 how much ADA they should deposit to qualify for the auction.
 However, none of this is enforced by any of the auction’s on-chain validators.
 
+Losing bidders can collect their deposit at any time after the bidding end time.
```
uhbif19: But Standing Bid can be cleaned after cleanup time and that means that `BidderDeposit` tx would not work.

test/EndToEnd/Ledger.hs
```diff
@@ -62,18 +64,34 @@ testSuite :: TestTree
...
     "L1"
-    [ testCase "bidder-buys" bidderBuysTest
+    [ testCase "bidder-buys" bidderBuysNoDepositTest
     , testCase "seller-reclaims" sellerReclaimsTest
     , testCase "seller-bids" sellerBidsTest
     , testCase "unauthorised-bidder" unauthorisedBidderTest
+    , testGroup "bidder-deposit" bidderDepositTests
     ]
 
+bidderDepositTests :: [TestTree]
```
uhbif19: Please separate them to another module. This one gets too big now, and you already separated new group.

test/EndToEnd/Ledger.hs
```diff
@@ -208,3 +226,250 @@ unauthorisedBidderTest = mkAssertion $ do
...
+
+losingBidderDoubleClaimTest :: Assertion
+losingBidderDoubleClaimTest = mkAssertion $ do
+  let seller = Alice
+      buyer1 = Bob
+      buyer2 = Carol
+
+  mapM_ (initWallet 100_000_000) [seller, buyer1, buyer2]
+
+  nftTx <- mintOneTestNFT
```
uhbif19: :thinking:  I think we should finally separate all that stuff.

uhbif19: Added to https://github.com/mlabs-haskell/hydra-auction/issues/267

test/EndToEnd/Ledger.hs
```diff
@@ -208,3 +226,250 @@ unauthorisedBidderTest = mkAssertion $ do
...
+  assertNFTNumEquals seller 0
+
+  withActor buyer1 $ newBid terms $ startingBid terms
+  withActor buyer2 $ newBid terms $ startingBid terms + minimumBidIncrement terms
+
+  waitUntil $ voucherExpiry terms
+
+  result <- try $ sellerClaimDepositFor terms (head actorsPkh)
+
+  case result of
```
uhbif19: I think that could be merged with `sellerClaimsDepositTest` cuz we have both winning and losing bidder here.

test/EndToEnd/Ledger.hs
```diff
@@ -208,3 +226,250 @@ unauthorisedBidderTest = mkAssertion $ do
...
+  assertUTxOsInScriptEquals Deposit terms 2
+
+  waitUntil $ biddingStart terms
+  actorsPkh <- liftIO $ getActorsPubKeyHash [buyer1, buyer2]
+  startBidding terms (ApprovedBidders actorsPkh)
+
+  assertNFTNumEquals seller 0
+
+  withActor buyer1 $ newBid terms $ startingBid terms
+  withActor buyer2 $ newBid terms $ startingBid terms + minimumBidIncrement terms
```
uhbif19: Actually we can separate all beginig and rename them to `losingBidder` and `winningBidder`, cuz that case seems to be reused everywhere.

src/HydraAuction/Tx/Deposit.hs
```diff
@@ -0,0 +1,211 @@
...
+          AutoCreateParams
+            { signedUtxos = [(bidderSk, bidderMoneyUtxo)]
+            , additionalSigners = []
+            , referenceUtxo = mempty
+            , witnessedUtxos = [(depositWitness, UTxO.singleton deposit)]
+            , collateral = Nothing
+            , outs = []
+            , toMint = TxMintValueNone
+            , changeAddress = bidderAddress
+            , validityBound = (Just $ cleanup terms, Nothing)
```
uhbif19: Could not this be defined by Stage?

src/HydraAuction/Tx/Deposit.hs
```diff
@@ -0,0 +1,211 @@
...
+  standingBidUtxo <- scriptUtxos StandingBid terms
+  allDeposits <- scriptUtxos Deposit terms
+
+  let mp = policy terms
+      voucherCS = VoucherCS $ scriptCurrencySymbol mp
+      expectedDatum = BidDepositDatum (toPlutusKeyHash $ verificationKeyHash bidderVk) voucherCS
+      depositScript = scriptPlutusScript Deposit terms
+      depositWitness = mkInlinedDatumScriptWitness depositScript LosingBidder
+
+  case UTxO.find ((== expectedDatum) . parseBidDepositDatum) allDeposits of
```
uhbif19: This stuff is reused everywhere, please separate to something like `findBidderDeposit`.

uhbif19 (changes requested):

src/HydraAuction/Tx/Deposit.hs
nazrhom: what do you mean?

test/EndToEnd/Ledger.hs
nazrhom: could we do this as part of #267? It's worth thinking of how to avoid all this repetition in the tests (maybe we could test the cli directly? and use stuff like prepare-for-demo)

test/EndToEnd/Ledger.hs
nazrhom: I generally prefer making tests very specific. For example, I noticed in the first tests we are always waiting for cleanup and eventually submitting the cleauptx, I think we should have a specific test for cleanup instead, and ignore it in the others. It does make tests a bit longer, but when anything breaks it is easier to understand exactly what it is.

src/HydraAuction/Tx/Deposit.hs
uhbif19: I thought we already have code for `validityBound` from stage generation. I was wrong.

test/EndToEnd/Ledger.hs
uhbif19: Okay.

My thinking was that checking everything separately in E2E tests may lead to case explosion.

hydra-auction.cabal
```diff
@@ -253,7 +253,8 @@ test-suite hydra-auction-test
   main-is:        Spec.hs
   other-modules:
     EndToEnd.Hydra
-    EndToEnd.Ledger
+    EndToEnd.Ledger.Auction
+    EndToEnd.Ledger.BidDeposit
```
uhbif19, line 266: Yay.

uhbif19 (approved):

PR #261, opened 2023-04-11
Delegate to hydra node implementation

### LINKED ISSUE #158: Implement Hydra node requests on delegate server

### LINKED ISSUE #174: Trigger automatic actions defined by Hydra events in delegate server
This task requires to introduce new thread placing in queue HydraEvents which are triggered by:

* `DelegateEvent HydraEvent`

For that, probably, some new utilities for Hydra responses monitoring are required.

Actual implementation of responses to them is out of scope of this task.

Blocked by #125
Blocked by https://github.com/mlabs-haskell/hydra-auction/issues/140

uhbif19:
CLI parsing is very hacked now, will fix that in following PR.

Closes #158 
Closes #174 

Blocked by #259

### DIFF
--- a/delegate-app/Main.hs
+++ b/delegate-app/Main.hs
@@ -352,9 +352,9 @@ main = do
   hydraNodeNumber <- read . fromJust <$> lookupEnv "HYDRA_NODE_NUMBER"
   putStrLn $ "With number: " <> show hydraNodeNumber
   let actor = case hydraNodeNumber of
-        1 -> Rupert
+        1 -> Oscar
         2 -> Patricia
-        3 -> Oscar
+        3 -> Rupert
         _ -> error "Wrong HYDRA_NODE_NUMBER"
   runHydraClientN hydraNodeNumber $ \hydraClient -> do
     liftIO $ putStrLn "Hydra connection opened"
--- a/delegate-app/Main.hs
+++ b/delegate-app/Main.hs
@@ -1,10 +1,12 @@
 module Main (main) where
 
 -- Prelude imports
-import Hydra.Prelude (lookupEnv, readMaybe, traverse_)
+-- Prelude imports
+import Hydra.Prelude (lookupEnv, putStr, readMaybe, traverse_)
 import Prelude
 
 -- Haskell imports
+
 import Control.Concurrent (newMVar, putMVar, takeMVar, threadDelay)
 import Control.Concurrent.Async (mapConcurrently_, race_)
 import Control.Concurrent.STM (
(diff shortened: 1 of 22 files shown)


delegate-app/Main.hs
```diff
@@ -165,28 +175,31 @@ runDelegateLogicSteps ::
...
-    flushAndTraverseNewElements
-      frontendRequestQueue
-      ( delegateFrontendRequestStep
-          >=> putInToClientsQueue
-      )
-    -- FIXME: log queues overload and make tick not-static
-    liftIO $ threadDelay tick
+  broadcast = do
+    (voidResult, _) <-
+      flip runStateT initialState $ forever $ do
```
nazrhom: nit: you can use `evalState` here to avoid having to pattern match on void an return it

delegate-app/Main.hs
```diff
@@ -306,23 +344,37 @@ mbQueueAuctionPhases delegateEvents toClientsChannel = do
...
-          { host
-          , port = fromMaybe portDefault $ port >>= readMaybe
-          , tick = tick
-          , ping = ping
-          }
-  runWithTracer' tracer $ runDelegateServer conf
+  -- FIXUP: parse actual adress and other params
+  hydraNodeNumber <- read . fromJust <$> lookupEnv "HYDRA_NODE_NUMBER"
+  putStrLn $ "With number: " <> show hydraNodeNumber
+  let actor = case hydraNodeNumber of
```
nazrhom, line 352: nit: I think it would be nicer to respect the natural ordering of the alphabet (oscar-patricia-rupert)

uhbif19: Yes, that is not only non-nice, but probably a reason for previous bug. I forgot to backport fix.

src/HydraAuction/OnChain/StandingBid.hs
```diff
@@ -76,28 +75,30 @@ mkStandingBidValidator terms datum redeemer context =
...
-          (oldApprovedBidders == newApprovedBidders)
-        && case oldBid of
-          Just oldBidTerms ->
-            traceIfFalse "Bid increment is not greater than minimumBidIncrement" $
-              bidAmount oldBidTerms + minimumBidIncrement terms <= bidAmount newBidTerms
-          Nothing ->
-            traceIfFalse "Bid is not greater than startingBid" $
-              startingBid terms <= bidAmount newBidTerms
+    validNewBid (StandingBidState _oldApprovedBidders oldBid) (StandingBidState _newApprovedBidders (Just newBidTerms)) =
+      -- FIXME: disabled until M6
```
nazrhom, line 79: Why do we need to remove this until m6? We should be able to make this work already

uhbif19: This breaks, cuz delegate is signing Tx, while it is not in list of bidders. And knowing delegate list by Frontend CLI also requires new code.

And in M6 solution will be changed to completely different anyway.

nazrhom (approved):

delegate-app/Main.hs
nazrhom: I noticed this is also wrong in the docker-compose, my bad :D

Issue #243, opened 2023-04-04
Change delegate server to receive hydra node args from command line rather than hardcoding them


nazrhom:
The delegate server needs some info on the hydra-node to talk to it (host/port). We currently hardcode these, but would be nice to use a similar interface to the CLI (optoparse-applicative) to have the delegate supply these parameters on startup

Cfr. https://github.com/mlabs-haskell/hydra-auction/pull/217#discussion_r1149790308

PR #263, opened 2023-04-11
Frontend CLI Delegate support draft

### LINKED ISSUE #205: Make Frontend CLI work with Delegate server
How CLI should work is covered by relevant section of off_chain_spec, while problems might occur.

1.

The first obstacle I see is tracking a delegate server/head state: probably it would be streamed by delegate server and should be recorded by listener thread. There is at least standing bid state and HeadId.

HeadId we should get from Delegate server before announcing auction. So CLI may probably just wait for it before starting or fail to do announce-auction command.

Current idea is to implement that is to update from `DelegateState` pushes from Delegate server. So that should probably be separate thread which waits for `DelegateState` and updates accordingly. That should be working the same in both Watch and Repl interfaces.

2.

Second issue is tracking L1/L2 status (and its possible interference with Auction stages). This should be tracked

I thing that we should introduce whole auction state state-machine datatype and store it in variable. It should keep track of both if standing bid is in L1 or L2 **and** the `DelegateState` from point 1 (obviously this is only applicable for L1 case).

So it would probably look like:

```
data AuctionState =
     NotAnnounced ...
     | OnL2 DelegateState
     | OnL1 ...
```

Tasks (draft):

- [x] L2 version of StartBidding action (see spec)
- [x] Tracking of L1/L2 status of bidding
- [x] Placing bid on L2 (see spec)
- [ ] Validate delegate state in CLI
- [ ] Cover possible problems with L1/L2 moving
- [ ] Changing CLI E2E test to check all of that.

Closing Head on L1 is out of scope of this task and in #160 

**UPD:** Separated E2E CLI tests to #239

Blocked by #175 
Blocked by #239
Blocked by #204

uhbif19:
Closes #205

Covered:

* Demo should work in theory

Not covered:

* [x] Error responses handling
* [x] Refactor TODOs, unsafe errors
* [x] Code duplication in new Actions
* [ ] New action: Start bidding on L2 (see spec)

Parse path to Delegate node as CLI arguments separated into #243 (but Mango done it anyway)
Handling `RequestIgnored` errors will be covered in #266 

Blocked by #261 

cc @MangoIV

### DIFF
--- a/app/CLI/Actions.hs
+++ b/app/CLI/Actions.hs
@@ -194,7 +194,6 @@ handleCliAction sendRequestToDelegate currentDelegateStateRef userAction = do
           delegateState <- liftIO $ readIORef currentDelegateStateRef
           case delegateState of
             Initialized headId _ -> do
-              liftIO $ putStrLn $ "HeadId is: " <> show headId
               dynamic <-
                 liftIO $
                   constructTermsDynamic
@@ -207,8 +206,6 @@ handleCliAction sendRequestToDelegate currentDelegateStateRef userAction = do
                   noteM ("could not read auction terms config for " <> show auctionName) $
                     readAuctionTermsConfig auctionName
               terms <- liftIO $ configToAuctionTerms config dynamic
-              -- TODO
-              liftIO . putStrLn $ show terms
               liftIO . putStrLn $
                 show actor
                   <> " announces auction called "
@@ -312,7 +309,6 @@ handleCliAction sendRequestToDelegate currentDelegateStateRef userAction = do
       utxos <- actorTipUtxo
       liftIO $ prettyPrintUtxo utxos
     Cleanup auctionName -> do
-      -- FIXME: proper error printing
       terms <- auctionTermsFor auctionName
       liftIO . putStrLn $
         "Cleaning up all remaining script utxos for auction "
--- a/app/CLI/Actions.hs
+++ b/app/CLI/Actions.hs
@@ -1,5 +1,6 @@
 module CLI.Actions (
   CliAction (..),
+  Layer (..),
   handleCliAction,
   seedAmount,
 ) where
@@ -74,7 +75,7 @@ import CLI.Prettyprinter (prettyPrintUtxo)
 seedAmount :: Lovelace
 seedAmount = 10_000_000_000
 
-data Layer = L1 | L2
+data Layer = L1 | L2 deriving stock (Show)
 
 data CliAction
   = ShowCurrentStage !AuctionName
--- a/app/CLI/Actions.hs
+++ b/app/CLI/Actions.hs
@@ -10,12 +10,15 @@ import Prelude
 
 -- Haskell imports
 import Control.Monad (forM_, void)
+import Data.IORef (IORef, readIORef)
 
 -- Plutus imports
 import Plutus.V1.Ledger.Address (pubKeyHashAddress)
 
 -- Hydra imports
+
 import Hydra.Cardano.Api (Lovelace, pattern ShelleyAddressInEra)
+import Hydra.Chain.Direct.Tx (headIdToCurrencySymbol)
 
 -- Cardano node imports
 import Cardano.Api.UTxO qualified as UTxO
@@ -45,8 +48,8 @@ import HydraAuction.Tx.Escrow (
   sellerReclaims,
   startBidding,
  )
-import HydraAuction.Tx.StandingBid (cleanupTx, currentWinningBidder, newBid)
-import HydraAuction.Tx.TermsConfig (constructTermsDynamic, nonExistentHeadIdStub)
+import HydraAuction.Tx.StandingBid (cleanupTx, createStandingBidDatum, currentWinningBidder, newBid)
+import HydraAuction.Tx.TermsConfig (constructTermsDynamic)
 import HydraAuction.Tx.TestNFT (findTestNFT, mintOneTestNFT)
 import HydraAuction.Types (ApprovedBidders (..), AuctionStage (..), AuctionTerms, BidDepositDatum (..), Natural, naturalToInt)
 import HydraAuctionUtils.Fixture (Actor (..), actorFromPkh, allActors, getActorsPubKeyHash)
@@ -63,6 +66,9 @@ import CLI.Config (
   writeAuctionTermsDynamic,
  )
 import CLI.Prettyprinter (prettyPrintUtxo)
+import Cardano.Api.UTxO qualified as UTxO
+import HydraAuction.Delegate.Interface (DelegateState (..))
+import HydraAuction.Delegate.Interface qualified as DelegateInterface
 
 seedAmount :: Lovelace
 seedAmount = 10_000_000_000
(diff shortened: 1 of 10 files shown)


app/CLI/Actions.hs
```diff
@@ -93,8 +102,13 @@ doOnMatchingStage terms requiredStage action = do
               <> show requiredStage
           )
 
-handleCliAction :: CliAction -> Runner ()
-handleCliAction userAction = do
+handleCliAction ::
+  (DelegateInterface.FrontendRequest -> IO ()) ->
+  IORef DelegateState ->
```
MangoIV, line 116: @uhbif19 is there a specific reason why we use an IORef here instead of a state transformer?

MangoIV:
I think I didn't quite understand what "proper CLI parsing" means here, afaict it's already parsing properly....

src/HydraAuction/Delegate/Interface.hs
```diff
@@ -77,10 +78,25 @@ data RequestIgnoredReason
   deriving stock (Eq, Show, Generic)
   deriving anyclass (FromJSON, ToJSON)
 
+instance Pretty RequestIgnoredReason where
+  pretty = \case
+    IncorrectData -> "data was incorrect"
+    WrongDelegateState state ->
+      vsep
+        [ "the wrong delegate was adressed"
```
uhbif19: This is incorrect.

src/HydraAuction/Delegate/Interface.hs
```diff
@@ -89,3 +105,28 @@ data DelegateResponse
...
+      vsep
+        [ "The request was ignored because"
+        , indent 2 $ pretty reason
+        ]
+    CurrentDelegateState reason state ->
+      "Delegate State Answer"
+        <> indent
+          2
+          ( vsep
+              [ "Responding because" <+> pretty reason
```
uhbif19, line 122: Why we need pretty printing all of this? This is debugging information. It should not be showed to user.

app/Main.hs
```diff
@@ -70,23 +72,29 @@ handleCliInput input = do
...
-        Just (CurrentDelegateState _ state) ->
+        Right (CurrentDelegateState _ state) ->
           writeIORef currentDelegateStateRef state
-        _ -> putStrLn $ "Ignoring server message: " <> show mMessage
+        Right response -> traceWith tracer (UnimplementedResponse response)
+        Left err -> traceWith tracer (InvalidResponse err)
 
-handleCliInput' :: _ -> IORef _ -> CliInput -> IO ()
-handleCliInput' client currentDelegateStateRef input = case input of
+handleQueryAction :: Connection -> IORef DelegateState -> CliInput -> IO ()
```
uhbif19: Why is it `QueryAction`?

delegate-app/Main.hs
```diff
@@ -349,7 +351,13 @@ main :: IO ()
 main = do
   port <- lookupEnv "PORT"
   -- FIXUP: parse actual adress and other params
-  hydraNodeNumber <- read . fromJust <$> lookupEnv "HYDRA_NODE_NUMBER"
+  envHydraNode <- lookupEnv "HYDRA_NODE_NUMBER"
```
uhbif19, line 356: There is FIXUP and PR todo which state that this is temporarily solution. It should be changed into proper CLI parser, not added errors.

app/CLI/Actions.hs
uhbif19: Because there is thread polling `DelegateState`.

src/HydraAuction/Delegate/Interface.hs
uhbif19: This error means that delegate in current `DelegateState` could not process request. Maybe my naming is not ideal.

src/HydraAuction/Delegate/Interface.hs
MangoIV: well it only shows with verbose, I thought that's what you meant with "errors" responses handling

app/Main.hs
MangoIV: I will change it back, I just changed it because it was easier to understand for me than `f`, `f'`, `f''`, ...

delegate-app/Main.hs
MangoIV: `read . fromJust` just errs with `error: fromJust`, which is very unhelpful, I just changed this because I wanted to know where the error comes from.

uhbif19:
>  I think I didn't quite understand what "proper CLI parsing" means here, afaict it's already parsing properly....

It uses env variables, instead of CLI args interface which is used for all other arguments.

>  Showing errors on not matching Delegate state

I meant `doOnMatchingStage` analog for DelegateState.

MangoIV:
thank you, that helped :)

uhbif19:
@MangoIV If you not started implementing CLI parsing, then I steal that part from you, cuz I already do same thing for Delegate server.

delegate-app/Main.hs
uhbif19: Okay.

uhbif19:
@MangoIV Updated TODO list. Will probably separate that to follow-up issue later.

uhbif19:
@MangoIV 

You checked "Refactor TODOs, unsafe errors and code duplication in new Actions" - but I did not see any changes at that part in your commit.

src/HydraAuction/Delegate/Interface.hs
MangoIV: changed

uhbif19:
@MangoIV Please rebase this PR over #261

app/CLI/Actions.hs
```diff
@@ -221,8 +222,10 @@ handleCliAction sendRequestToDelegate currentDelegateStateRef userAction = do
               , utxoToCommit = standingBidTxIn
               }
     NewBid auctionName bidAmount -> do
-      -- FIXME: proper error printing
-      Just CliEnhancedAuctionTerms {terms, sellerActor} <- liftIO $ readCliEnhancedAuctionTerms auctionName
+      CliEnhancedAuctionTerms {terms, sellerActor} <-
```
uhbif19, line 241: L1/L2 actions are still essentially copy-pasted which stands for `code duplication in new Actions`.

app/CLI/Actions.hs
```diff
@@ -303,10 +306,19 @@ handleCliAction sendRequestToDelegate currentDelegateStateRef userAction = do
       liftIO $ prettyPrintUtxo utxos
     Cleanup auctionName -> do
       -- FIXME: proper error printing
-      Just terms <- liftIO $ readAuctionTerms auctionName
+      terms <- auctionTermsFor auctionName
```
uhbif19, line 313: Thanks.

app/CLI/Watch.hs
```diff
@@ -3,7 +3,8 @@ module CLI.Watch (
 ) where
 
 -- Prelude
-import Hydra.Prelude (MonadDelay (threadDelay))
+-- Prelude
```
uhbif19:
```suggestion
```

app/CLI/Types.hs
```diff
@@ -0,0 +1,25 @@
...
+import HydraAuction.Delegate.Interface (DelegateResponse)
+import Prettyprinter (Doc, Pretty (pretty), indent, line, (<+>))
+import Prelude
+
+newtype CLILog
+  = CLIError CLIError
+
+data CLIError
+  = UnimplementedResponse DelegateResponse
+  | InvalidResponse String
```
uhbif19:
```suggestion
  | InvalidDelegateResponse String
```

uhbif19 (commented):
I cannot approve, cuz I created PR, but lgtm. I updated TODO.

uhbif19 (commented):
I cannot approve, cuz I created PR, but lgtm. I updated TODO.

MangoIV (approved):
approval by @uhbif19

Issue #268, opened 2023-04-13
Support for off-chain validation and intervals for `AuctionStage` and fix boundary case


uhbif19:
Couple of Tx do not fall into single stage. This requires hardcoding of intervals. Support for intervals over `AuctionStage` would cover this.

Also currently intervals for `Tx.*` are created manually.

- [x] Fix boundary case (@uhbif19 has draft for that locally)
- [ ] Support intervals of stages (or maybe just stage-to-infinity case, which is only used in our code)
- [x] Support Tx `validityBound` construction
- [x] Refactor current code to use this

PR #274, opened 2023-04-21
Utility improvements and fixes in preparation for Delegate corner-cases implementation

uhbif19:
Non trivial feature: https://github.com/mlabs-haskell/hydra-auction/pull/274/commits/d914f37b302d3ad82836596cce8dd081164fc245

I think all other commits should be trivial.

Blocked by #272  (cherry-picked)
Blocked by #271


src/HydraAuction/OnChain/StandingBid.hs
```diff
@@ -121,3 +117,15 @@ mkStandingBidValidator terms datum redeemer context =
...
+validNewBidTerms :: AuctionTerms -> Maybe BidTerms -> Maybe BidTerms -> Bool
+validNewBidTerms terms oldBid (Just newBidTerms) =
+  case oldBid of
+    Just oldBidTerms ->
+      traceIfFalse "Bid increment is not greater than minimumBidIncrement" $
+        bidAmount oldBidTerms + minimumBidIncrement terms <= bidAmount newBidTerms
+    Nothing ->
+      traceIfFalse "Bid is not greater than startingBid" $
+        startingBid terms <= bidAmount newBidTerms
+validNewBidTerms _14 _ Nothing =
```
nazrhom:
```suggestion
validNewBidTerms _ _ Nothing =
```

src/HydraAuction/Tx/Common.hs
```diff
@@ -121,7 +131,9 @@ currentAuctionStage terms = do
   let matchingStages = filter (member currentTime . stageToInterval terms) auctionStages
   return $ case matchingStages of
     [stage] -> stage
-    _ -> error "Impossible happend: more than one matching stage"
+    [_, stage] -> stage
```
nazrhom, line 134: Why do we need this?

uhbif19: Because this code fails (like I actually saw it). As I understand in case of stages corner.

While my code is not good - we need test on this.

uhbif19: Or maybe this is just wrong and intervals should be ensured to be not intersecting? I missed that.

nazrhom: I thought that was the idea, if this fails it means we must have an overlap in the stages

uhbif19:
@nazrhom I returned `[seller, bidder1, bidder2] <- return [Alice, Bob, Carol]` stuff back, otherwise GHC produces warning.

src/HydraAuction/Tx/Common.hs
uhbif19: Simple fix breaks Tx construction, cuz they use Upper bounds too, so I added that task into: https://github.com/mlabs-haskell/hydra-auction/issues/268

nazrhom (approved):

Issue #256, opened 2023-04-09
Integrate bidder/seller deposit claim to Frontend CLI


uhbif19:
Todo:

- [x] Add winning bidder claim to Frontend CLI (to `BidderBuys` action)
- [x] Add seller/losing bidder claims to Frontend CLI (TBD how)

Blocked by #246 
Blocked by #257

PR #285, opened 2023-04-26
Add BidderClaimsLosingDeposit and SellerClaims deposit to CLI. Also fix CLI tests as they had broken in some merge

nazrhom:


nazrhom:
Partially resolves #256

uhbif19:
@nazrhom Please separate to commits.

hydra-auction.cabal
```diff
@@ -307,5 +311,5 @@ test-suite hydra-auction-test
     , tasty-hunit
     , text
 
-  hs-source-dirs: test
+  hs-source-dirs: test app
```
uhbif19, line 320: "/

I think we should separate CLI to another lib or something like that.

test/EndToEnd/CLI.hs
```diff
@@ -1,118 +1,145 @@
...
+      fn <- toJsonFileName AuctionConfig auctionName
+      removeFile fn
+
+mockDelegateState :: DelegateState
+mockDelegateState = Initialized (HeadId . fromBuiltin . unCurrencySymbol $ nonExistentHeadIdStub) (Open Nothing)
+
+auctionName :: AuctionName
+auctionName = "test"
+
+handleCliActionWithMockDelegates :: CliAction -> Runner ()
```
uhbif19: Nice! I think L2 could be tested too with similar helper.

test/EndToEnd/CLI.hs
```diff
@@ -1,118 +1,145 @@
...
-    }
+  withResource createTestConfig deleteTestConfig $
+    const $
+      testGroup
+        "L1 - CLI"
+        [ testCase "bidder-buys" bidderBuysTest
+        , testCase "deposit-test" depositTest
+        ]
+  where
+    createTestConfig = writeAuctionTermsConfig auctionName config
```
uhbif19, line 55: I think this could break real configs, no?

test/EndToEnd/CLI.hs
```diff
@@ -1,118 +1,145 @@
...
+
+  assertNFTNumEquals seller 0
+  assertNFTNumEquals buyer1 0
+  assertNFTNumEquals buyer2 1
+
+  assertUTxOsInScriptEquals Deposit terms 1
+
+  withActor buyer1 $ handleCliActionWithMockDelegates $ BidderClaimsDeposit auctionName
+
+  assertUTxOsInScriptEquals Deposit terms 0
```
uhbif19, line 143: Would you please check actor Utxos too? Or better - introduce real ADA checks.

uhbif19 (approved):
Sorry, forgot to push review yesterday.

test/EndToEnd/CLI.hs
nazrhom: @uhbif19 you are right... Users of the cli will not be running tests, but it is a bit annoying anyway.... Maybe we could generate a random uuid for the test name or something

test/EndToEnd/CLI.hs
nazrhom: Real ada checks are a bit tricky, because its hard to track fees, so we don't really have precise numbers to check.

Checking actor utxos would work (we will expect every losing bidder to have 1 utxo, as the claiming consolidates everything, and the winning bidder to have 2), but it is not a great test, as it would succeed even if the `BidderClaimsDeposit` tx sends the funds to the wrong account

uhbif19: Fees can be taken from Tx or TxBody and taken into account.

Issue #267, opened 2023-04-13
Refactor EndToEnd tests


uhbif19:
- [x] Separate common preparations in EndToEnd tests
- [x] Fix code duplication between EndToEnd tests
- [x] Remove `EndToEnd.Hydra` tests

Blocked by #248

PR #297, opened 2023-05-02
Nazrhom/fee escrow

### LINKED ISSUE #255: Implement fee escrow on-chain and tx constuction
Details are in `on_chain_spec`. 

- [x] Fee escrow: on chain validator, distribution tx construction and tests for it
- [x] Require fee escrow output in escrow outputs - on chain, tx construction and tests

nazrhom:
Part of #267

### DIFF
--- a/src/HydraAuction/OnChain/FeeEscrow.hs
+++ b/src/HydraAuction/OnChain/FeeEscrow.hs
@@ -1,0 +1,62 @@
+{-# OPTIONS_GHC -fno-specialise #-}
+
+module HydraAuction.OnChain.FeeEscrow (mkFeeEscrowValidator) where
+
+-- Prelude imports
+import PlutusTx.Prelude
+import Prelude (maximum, minimum)
+
+-- Plutus imports
+
+import Plutus.V1.Ledger.Value (valueOf)
+import Plutus.V2.Ledger.Api (ScriptContext, TxInfo (..), TxOut (..), Value, adaSymbol, adaToken, scriptContextTxInfo, txInInfoResolved)
+
+-- Hydra auction imports
+import HydraAuction.Types (
+  AuctionTerms (..),
+  FeeEscrowDatum,
+  FeeEscrowRedeemer (..),
+  )
+import HydraAuctionUtils.Types.Natural (naturalToInt)
+import HydraAuctionUtils.Plutus (
+  nothingForged,
+ )
+
+{-# INLINEABLE mkFeeEscrowValidator #-}
+mkFeeEscrowValidator :: AuctionTerms -> FeeEscrowDatum -> FeeEscrowRedeemer -> ScriptContext -> Bool
+mkFeeEscrowValidator terms () DistributeFees context =
+  -- There is one output per delegate. The conditions in validFeeDistribution are satisfied when applied to these outputs and the transaction fee.
+  length outputs == length (delegates terms)
+    && validFeeDistribution outputs singleFeeInputValue
+    -- No tokens are minted or burned.
+    && nothingForged info
+  where
+    info :: TxInfo
+    info = scriptContextTxInfo context
+
+    outputs :: [TxOut]
+    outputs = txInfoOutputs info
+
+    validFeeDistribution :: [TxOut] -> Value -> Bool
+    validFeeDistribution outputsToDelegates txFee =
+      allAdaDistributed
+        && adaDistributedEvenly
+      where
+        -- Each delegate received the `auctionFeePerDelegate`,
+        -- after deducting the transaction fees from the total.
+        allAdaDistributed = actualTotalAda == expectedTotalAda
+        actualTotalAda = sum actualAdaValues + adaValueOf txFee
+        expectedTotalAda = length (delegates terms) * naturalToInt (auctionFeePerDelegate terms)
+
+        -- The amount received by any delegate differs by at most one lovelace
+        -- from what any other delegate received.
+        adaDistributedEvenly = True -- 1 > (maximum actualAdaValues - minimum actualAdaValues)
+        adaValueOf v = valueOf v adaSymbol adaToken
+        actualAdaValues = adaValueOf . txOutValue <$> outputsToDelegates
+
+    -- There is one input spent from the fee escrow validator.
+    singleFeeInputValue :: Value
+    singleFeeInputValue = case txInInfoResolved <$> txInfoInputs info of
+      [] -> traceError "Missing input for fee escrow"
+      [feeOut] -> txOutValue feeOut
+      _ : _ -> traceError "More than single input from fee escrow validator"
--- a/src/HydraAuction/OnChain/FeeEscrow.hs
+++ b/src/HydraAuction/OnChain/FeeEscrow.hs
@@ -4,30 +4,31 @@ module HydraAuction.OnChain.FeeEscrow (mkFeeEscrowValidator) where
 
 -- Prelude imports
 import PlutusTx.Prelude
-import Prelude (maximum, minimum)
+
+-- import Prelude (quot)
 
 -- Plutus imports
-
+import Plutus.V1.Ledger.Address (pubKeyHashAddress, scriptHashAddress)
 import Plutus.V1.Ledger.Value (valueOf)
-import Plutus.V2.Ledger.Api (ScriptContext, TxInfo (..), TxOut (..), Value, adaSymbol, adaToken, scriptContextTxInfo, txInInfoResolved)
+import Plutus.V2.Ledger.Api (TxInfo (..), TxOut (..), Value, adaSymbol, adaToken, scriptContextTxInfo, txInInfoResolved)
+import Plutus.V2.Ledger.Contexts (ScriptContext, ownHash)
 
 -- Hydra auction imports
 import HydraAuction.Types (
   AuctionTerms (..),
   FeeEscrowDatum,
   FeeEscrowRedeemer (..),
-  )
-import HydraAuctionUtils.Types.Natural (naturalToInt)
+ )
 import HydraAuctionUtils.Plutus (
+  byAddress,
   nothingForged,
  )
 
 {-# INLINEABLE mkFeeEscrowValidator #-}
 mkFeeEscrowValidator :: AuctionTerms -> FeeEscrowDatum -> FeeEscrowRedeemer -> ScriptContext -> Bool
 mkFeeEscrowValidator terms () DistributeFees context =
-  -- There is one output per delegate. The conditions in validFeeDistribution are satisfied when applied to these outputs and the transaction fee.
-  length outputs == length (delegates terms)
-    && validFeeDistribution outputs singleFeeInputValue
+  -- Every delegate is payed at least the expected proportion of fee
+  all receivesProportionOfFee (delegates terms)
     -- No tokens are minted or burned.
     && nothingForged info
   where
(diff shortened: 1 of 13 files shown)


src/HydraAuction/OnChain/FeeEscrow.hs
```diff
@@ -0,0 +1,57 @@
...
+  FeeEscrowDatum,
+  FeeEscrowRedeemer (..),
+ )
+import HydraAuctionUtils.Plutus (
+  byAddress,
+  nothingForged,
+ )
+
+{-# INLINEABLE mkFeeEscrowValidator #-}
+mkFeeEscrowValidator :: AuctionTerms -> FeeEscrowDatum -> FeeEscrowRedeemer -> ScriptContext -> Bool
```
nazrhom, line 29: This is implemented slightly differently as discussed on the spec, as the suggested approach has a few complications. I'll raise this in the meeting today, and will update the spec as well if the changes I propose make sense

src/HydraAuction/Tx/TermsConfig.hs
```diff
@@ -67,8 +69,8 @@ instance ToJSON AuctionTermsDynamic
 instance FromJSON AuctionTermsDynamic
 
 -- | Stub for tests not checking MoveToHyda. Something not existent.
-nonExistentHeadIdStub :: CurrencySymbol
-nonExistentHeadIdStub = "DEADBEEF"
+nonExistentHeadIdStub :: HeadId
```
uhbif19, line 73: Thanks :+1:

test/EndToEnd/Utils.hs
```diff
@@ -46,12 +46,12 @@ config =
     , configDiffVoucherExpiry = 8
     , configDiffCleanup = 10
     , configAuctionFeePerDelegate = fromJust $ intToNatural 4_000_000
-    , configStartingBid = fromJust $ intToNatural 8_000_000
-    , configMinimumBidIncrement = fromJust $ intToNatural 8_000_000
+    , configStartingBid = fromJust $ intToNatural 15_000_000
```
uhbif19, line 49: Why?

src/HydraAuction/OnChain/FeeEscrow.hs
```diff
@@ -0,0 +1,57 @@
+{-# OPTIONS_GHC -fno-specialise #-}
+
+module HydraAuction.OnChain.FeeEscrow (mkFeeEscrowValidator) where
+
+-- Prelude imports
+import PlutusTx.Prelude
+
+-- import Prelude (quot)
```
uhbif19, line 8:
```suggestion
```

src/HydraAuction/Tx/FeeEscrow.hs
```diff
@@ -0,0 +1,118 @@
...
+  logMsg "Distributing fees to delegates"
+
+  (actorAddress, _, actorSk) <- addressAndKeys
+
+  actorMoneyUtxo <- filterAdaOnlyUtxo <$> actorTipUtxo
+
+  feeEscrowUtxo <- scriptUtxos FeeEscrow terms
+
+  feeEscrowValue <- case UTxO.pairs feeEscrowUtxo of
+    [(_, txOut)] -> pure $ txOutValue txOut
```
uhbif19, line 72: There is a `scriptSingleUtxo` function for that.

src/HydraAuction/OnChain/FeeEscrow.hs
```diff
@@ -0,0 +1,57 @@
...
+    info :: TxInfo
+    info = scriptContextTxInfo context
+
+    outputs :: [TxOut]
+    outputs = txInfoOutputs info
+
+    adaValueOf v = valueOf v adaSymbol adaToken
+
+    expectedValuePerDelegate :: Integer
+    expectedValuePerDelegate = adaValueOf singleFeeInputValue `quotient` length (delegates terms)
```
uhbif19: This is just `auctionFeePerDelegate`. This cannot have a reminder, cuz `Escrow` ensures that `calculateTotalFee terms` is emitted. If it was not so amount should be checked anyway.

 So no need for all this stuff.

uhbif19 (changes requested):

src/HydraAuction/Tx/FeeEscrow.hs
```diff
@@ -0,0 +1,118 @@
...
+zipDistributingValue lovelaceAmt delegatePKHs = zip delegatePKHs (lovelaceToValue . Lovelace <$> distributedLovelace)
+  where
+    delegateNumber = length delegatePKHs
+    (quotient, remainder) = divMod lovelaceAmt (toInteger delegateNumber)
+
+    spread _ [] = []
+    spread 0 xs = xs
+    spread k (x : xs) = (x + 1) : spread (k - 1) xs
+
+    distributedLovelace = spread remainder (replicate delegateNumber quotient)
```
uhbif19, line 117: This is not needed because of `auctionFeePerDelegate` too

uhbif19: Or is all this code due to #298 ? In that case please put comments for `spread` function and write about this motivation in spec.

uhbif19 (approved):

src/HydraAuction/Tx/FeeEscrow.hs
```diff
@@ -0,0 +1,118 @@
...
+    mkFeeOut (delegatePKH, valuePerDelegate) = do
+      delegateAddress <- fromPlutusAddressInMonad $ pubKeyHashAddress delegatePKH
+      pure $
+        TxOut
+          delegateAddress
+          valuePerDelegate
+          TxOutDatumNone
+          ReferenceScriptNone
+
+zipDistributingValue :: Integer -> [PubKeyHash] -> [(PubKeyHash, Value)]
```
uhbif19:
```suggestion
zipDistributingValue :: Lovelace -> [PubKeyHash] -> [(PubKeyHash, Value)]
```

src/HydraAuction/Tx/FeeEscrow.hs
uhbif19: If we leave this kind of logic, it should have property-based unit test matching it with on-chain check.

So I think it may be simpler to just reuse this function in on-chain too. That way there is no need to test that they match each over. And it seems like there is no use for indeterminacy of remainder spreading anyway.

While this needs unit-test or property-based test anyway.

uhbif19 (changes requested):
I now noticed that now this code actually uses delegates list. But there is not way to actually get it now. It should be returned by delegate server I gues.

src/HydraAuction/Tx/FeeEscrow.hs
nazrhom: > Or is all this code due to https://github.com/mlabs-haskell/hydra-auction/issues/298 ? I

No, this is only distributing ADA to the delegates in a balanced way.

> If we leave this kind of logic, it should have property-based unit test matching it with on-chain check.

> So I think it may be simpler to just reuse this function in on-chain too. That way there is no need to test that they match each over. And it seems like there is no use for indeterminacy of remainder spreading anyway.

> While this needs unit-test or property-based test anyway.

I don't think we really need unit tests for this helper function, its quite a trivial implementation, we already have on and offchain tests for this anyway, so we are testing it.

Also, I don't think it would be nice to use this on-chain, this gives slightly more ADA to the delegates appearing first in the list. In the on-chain we should only verify that each delegates receives >= `auctionFeePerDelegate`.
The off-chain has to actually consume the utxo, so in case there is more ada than `auctionFeePerDelegate * length delegates` it can decide how to distribute the extra ada

nazrhom:
@uhbif19 this is currently hardcoded to be `hydraNodeActors` in https://github.com/mlabs-haskell/hydra-auction/pull/297/files#diff-845d2a2cc5430818b2ff390b4bac10a80f1ebb2ada8af17b27e08fd915b82deaR84

src/HydraAuction/Tx/FeeEscrow.hs
nazrhom: I don't think it work as nicely in this case, as I want both the reference to `feeEscrowUtxo` and to look at its value

src/HydraAuction/Tx/FeeEscrow.hs
uhbif19: > No, this is only distributing ADA to the delegates in a balanced way.

I mean that if we do are not concerned with #298 then this ADA separation code is not needed and just `auctionFeePerDelegate` can be used.

> I don't think we really need unit tests for this helper function, its quite a trivial implementation

Integer division and such logic is exactly the kind of stuff where people get errors like off-by-one or something like that. Also we already got off-by-one errors in intervals helpers code.

> we already have on and offchain tests for this anyway, so we are testing it.

Our tests do not test ADA amount at all and this is an accident waiting to be happen.

> this gives slightly more ADA to the delegates appearing first in the list

Yes, and your specification says that this is okay. So why accept different remainder distributions then if spec says that any of them are good?

> The off-chain has to actually consume the utxo, so in case there is more ada than auctionFeePerDelegate * length delegates it can decide how to distribute the extra ada

What is `off-chain`? Why where can be extra ADA? If it is winning lot, then this does not interfere with delegate ADA distribution, cuz bidder cannot be an delegate.

uhbif19: @GeorgeFlerovsky Can delegate be a bidder too?

uhbif19:
> this is currently hardcoded to be hydraNodeActors

I understand. I meant real CLI. We can separate CLI fixing for that stuff in another issue, which probably requires delegates list returining, and merge this.

src/HydraAuction/Tx/FeeEscrow.hs
```diff
@@ -105,8 +104,8 @@ distributeFee terms = do
           TxOutDatumNone
           ReferenceScriptNone
 
-zipDistributingValue :: Integer -> [PubKeyHash] -> [(PubKeyHash, Value)]
-zipDistributingValue lovelaceAmt delegatePKHs = zip delegatePKHs (lovelaceToValue . Lovelace <$> distributedLovelace)
```
uhbif19: Lovelace seems to implement `divMod`, so (un)wrapping is not really necessary.

image

src/HydraAuction/OnChain/FeeEscrow.hs
```diff
@@ -23,12 +23,15 @@ import HydraAuctionUtils.Plutus (
...
  )
+import HydraAuctionUtils.Types.Natural (naturalToInt)
 
 {-# INLINEABLE mkFeeEscrowValidator #-}
 mkFeeEscrowValidator :: AuctionTerms -> FeeEscrowDatum -> FeeEscrowRedeemer -> ScriptContext -> Bool
 mkFeeEscrowValidator terms () DistributeFees context =
-  -- Every delegate is payed at least the expected proportion of fee
-  all receivesProportionOfFee (delegates terms)
+  -- There is one input spent from the fee escrow validator with enough ADA to pay the fees
+  adaValueOf singleFeeInputValue >= expectedFeePerDelegeate * length (delegates terms)
```
uhbif19, line 32: But if all delegates get their fee, then everything is okay by construction, no?

uhbif19 (approved):

uhbif19 (approved):

src/HydraAuction/Tx/FeeEscrow.hs
uhbif19: It returns `Maybe (TxIn, TxOut CtxUTxO)` which seems the same that you use.

Probably the name `..Utxo` is confusing, but I still do not found how to name `(TxIn, TxOut)` pair better. `Utxo` term being overloaded is awful, and fixing better vocabluary would be cool, but I do not know how to name it better.

cc @GeorgeFlerovsky

PR #302, opened 2023-05-08
Add spec for bidder signed datums and open auctions

nazrhom:


doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
...
+a pair consisting of the `hydraHeadId` and `bidderPKH`
+
+```haskell
+import Crypto.Sign.Ed25519
+
+generateSellerSignature :: SigningKey PaymentKey -> HeadId -> PubKeyHash -> BuiltinByteString
+generateSellerSignature skey headId bidderPKH = toBuiltin $ dsign skey $ sellerSignatureMessage headId bidderPKH 
+
+sellerSignatureMessage :: HeadId -> PubKeyHash -> BuiltinByteString
+sellerSignatureMessage headId bidderPKH = toByteString headId <> toBysteString bidderPKH
```
GeorgeFlerovsky: Not `headID`. Auction ID (i.e. the asset class / currency symbol of the voucher token)

doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
 can be done in a single transaction.
 Later on, this can be generalized in a straightforward way
 to accommodate incremental fee distribution to a larger number of delegates.
+
+# Extensions
+
+## Open Auctions
+
+This section describes an extenion that transforms the auctions from being closed to open.
```
GeorgeFlerovsky: No, "open auction" means that anyone can start bidding without seller authorization. That is outside the scope of our current project.

The seller signatures feature merely upgrades the closed auction from static membership set at bidding start to dynamic membership — whereby the seller can authorize bidders via signature even after bidding starts.

doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
...
+  -- ^ PubKeyHash of whoever submitted the bid
+  , bidderVK :: BuiltinByteString
+  -- ^ New field. Verification Key of whoever submitted the bid
+  , bidPrice :: Natural
+  -- ^ Which price did the bidder set to buy the auction lot?
+  , bidderSignature :: BuiltinByteString
+  -- ^ New field. Represents the signed payload by the bidder
+  , sellerSignature :: BuiltinByteString
+  -- ^ Represents the signed payload by the seller
+  }
```
GeorgeFlerovsky, line 795: Needs to mention the auction's voucher asset class. The bidder is bidding in a specific auction, not just announcing a number to the heavens. 😄

doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
...
+```
+
+This signature can be generated by the bidder,
+by signing with their signing key,
+a tuple consisting of the `hydraHeadId`, `bidderPKH` and `bidAmount`
```
GeorgeFlerovsky, line 800: Not the hydra head ID, but the voucher asset class.

GeorgeFlerovsky (changes requested):
I think there's been some confusion regarding the term "open auction" and "auction ID".

I added some comments to clarify.

doc/on_chain_spec.md
uhbif19: I suppose that @nazrhom change presumes that we change this behavior now. In that case this should not be `extension` but just spec change.

Also deprecated `ApprovedBidders` stuff should be removed from spec, cuz it is already confusing and obsolete.

doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
...
+```
+
+We also need to extend the previous implementation of `sellerSignatureMessage` to also include the bidder verification key in the signature.
+We will use this to verify the association between bidder PKH and VK on-chain.
+
+```haskell
+generateSellerSignature :: SigningKey PaymentKey -> HeadId -> PubKeyHash -> SigningKey VerificationKey ->  BuiltinByteString
+generateSellerSignature skey = toBuiltin . (dsign skey) . sellerSignatureMessage
+
+sellerSignatureMessage :: HeadId -> PubKeyHash -> SigningKey VerificationKey -> BuiltinByteString
```
uhbif19: Maybe `bidderApprovalMessageToSign`?

uhbif19: There is no such term as "signature message", I think.

doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
...
+import Crypto.Sign.Ed25519
+
+generateSellerSignature :: SigningKey PaymentKey -> HeadId -> PubKeyHash -> BuiltinByteString
+generateSellerSignature skey headId bidderPKH = toBuiltin $ dsign skey $ sellerSignatureMessage headId bidderPKH 
+
+sellerSignatureMessage :: HeadId -> PubKeyHash -> BuiltinByteString
+sellerSignatureMessage headId bidderPKH = toByteString headId <> toBysteString bidderPKH
+```
+
+We can now extend `BidTerms` to contain this new `sellerSignature`,
```
uhbif19, line 699: `sellerSignature` - I think this should be named `bidderParticipationApproval` or something like that. `sellerSignature` does not explain what this signature does and what is it for.

doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
...
+
+
+## Signed bidder datums
+
+This section describes an extension to the above spec,
+such that the standing bid is extended with a signed payload by the bidder.
+
+It assumes the `Open Auctions` extension to be implemented
+as it relies on the seller
+for enstablishing the on-chain link between bidder verification key and pubkey hash
```
uhbif19:
```suggestion
for establishing the on-chain link between bidder verification key and pubkey hash
```

uhbif19: > enstablishing the on-chain link between bidder verification key and pubkey hash

I think that could be named just `pubkey hash oracle`.  Not sure that term exist, "enstablishing the on-chain link" is what oracle does.

doc/on_chain_spec.md
uhbif19: And all those repeating args are code smell. Could them be just `BidderAprovalMessage` datatype?

Also that way we could use is as phantom type to parameterize `BuiltinByteString`.

doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
...
+  -> StandingBidState
+  -> StandingBidState
+  -> Bool
+validNewBid AuctionTerms{..} oldStandingBidState newStandingBidState =
+  isJust $ do
+    let StandingBidState oldApprovedBidders oldBid = oldStandingBidState
+    let StandingBidState newApprovedBidders newBid = newStandingBidState
+    BidTerms bidderPKH bidderVK newPrice bidderSignature sellerSignature <- newBid
+    -- The seller has allowed the bidder to participate in the auction
+    guard $ verifyEd25519Signature (sellerVK terms) (sellerSignatureMessage (hydraHeadId terms) bidderPKH bidderVK) sellerSignature
```
uhbif19: This should be checked by bidder cli as well to prevent wrong PKH forging by seller :D

That means:

1. Off-chain should be changed as well
2. This check should be separated for reuse

doc/on_chain_spec.md
```diff
@@ -669,3 +669,175 @@ that distributing their respective portions of the auction fee
...
+    -- The bidder has correctly signed the datum
+    guard $ verifyEd25519Signature bidderVK (bidderSignatureMessage (hydraHeadId terms) newPrice bidderPKH) bidderSignature
```
uhbif19: Was not the whole point of `bidderVK` link oracle solution to prevent this check and use `txSignedBy`?

If you check it like this, then `bidderVK/PKH` link seems to be unused. So it could just be dropped as I proposed in Slack.

uhbif19 (changes requested):

doc/on_chain_spec.md
nazrhom: I think the term oracle might be slightly misleading, it makes me think the seller would publish something on-chain to guarantee the relationship, but in this case the signature is only passed around off-chain.

I'll add a mention to the concept, but maybe try to clarify this difference

doc/on_chain_spec.md
nazrhom: I am using the `headId` as unique identifier of the auction, would that not work?

doc/on_chain_spec.md
nazrhom: Not necessarily, the point is to establish the link between VK and PKH on-chain. This way we can keep using the PKH "downstream". If we drop the PKH completely, then we need to introduce on-chain verification in a bunch of other places (off the top of my head: at least in bidder-buys and losing-bidder-claims-deposit, maybe also somewhere else?)

doc/on_chain_spec.md
nazrhom: We could do this, but I don't know how much sense it makes to use an on-chain primitive to verify something off-chain.

The CLI already needs to import some crypto library to build the bidder signature, so it might as well just use the functions it offers to verify them as well

doc/on_chain_spec.md
uhbif19: You right, I did not think about "downstream" txs.

doc/on_chain_spec.md
nazrhom: Sorry, but I really disagree on this point. I think we should strive to make this project as simple to understand as possible, it is meant to be a reference implementation for others to pickup.

What is the benefit of making this a separate datatype? All I really need, is to define in a single place where a message is constructed, this way it can be easily changed around if needed, and also shared by on and off chain.

The idea of using something like a phantom type to tag the kind of messages we are building is nice, but it also comes at the cost of making the code more complicated. If we were signing a bunch of different messages all over the place it could make sense, but in our case I don't think it is worth the extra complexity.

doc/on_chain_spec.md
uhbif19: HeadIds for started Heads should be unique. But as field in AuctionTerms it has no validation, until it got into Head. So announced, not commited yet, auction could use any HeadId.

While voucher asset is guarantied to be unique.

doc/on_chain_spec.md
uhbif19: > but in this case the signature is only passed around off-chain

Well it meant to be used on-chain eventually.

But you right, probably term "oracle" is reserved for publishing stuff on-chain in advance. Then I do not know common terms for what we do here :)

doc/on_chain_spec.md
uhbif19: > use an on-chain primitive to verify something off-chain

I do not understand what you mean. I was talking about re-checking it off-chain, but on client side.

doc/on_chain_spec.md
nazrhom: Ahhh makes sense, thanks!

doc/on_chain_spec.md
nazrhom: Sorry, I was replying to this point specifically:

> This check should be separated for reuse

I agree with everything else you said.

What i meant is that `verifyEd25519Signature` is part of plutustx, the client side will already need to import something like `Crypto.Sign.Ed25519` to build the bidder signature, so it might as well use the `verify` function from there to to the check you mentioned

nazrhom:
Closing in favour of #308

PR #307, opened 2023-05-16
Same delegate server case and rebidding by server case

uhbif19:
Blocked by https://github.com/mlabs-haskell/hydra-auction/pull/305

### DIFF
--- a/src/HydraAuction/OnChain.hs
+++ b/src/HydraAuction/OnChain.hs
@@ -19,8 +19,9 @@ import PlutusTx.Prelude
 import Prelude qualified
 
 -- Plutus imports
-import Plutus.V1.Ledger.Value (AssetClass (..))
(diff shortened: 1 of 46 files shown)


src/HydraAuction/OnChain.hs
```diff
@@ -118,10 +119,11 @@ feeEscrowAddress = FeeEscrowAddress . validatorAddress . feeEscrowValidator
 
 -- Deposit
 
+-- HACK: compilies incorrectly without HasCallStack
```
nazrhom, line 122: This should not be the fix. This has most likely something do to with inlining, or specialisation, as adding this random constraint seems to fix the issue.

uhbif19: Yes, but I already tried to introduce PRAGMAs everywhere. Also this kind of fail is not mentioned on troubleshooting page and I did not found it in issues.

I plan to ask in chat if anyone knows about it.

nazrhom (approved):

PR #308, opened 2023-05-22
Nazrhom/m6 impl

### LINKED ISSUE #203: Add bidder authentication in NewBid transactions

### LINKED ISSUE #220: Write on-chain `Integer -> BuiltinBytesting` serialization
- [ ] Write protorype how can we quickly serialize numbers -> bytesting onchain and back
- [ ] Check that this does this fit in one transaction
- [x] Ask koz, because he should have done it according to the issue he opened wrt supporting it natively 

This question is raised by #203

### LINKED ISSUE #221: Check bidder auction approval by seller signed datum
(such stuff are probably named "token" as in JWT, but word "token" already have a different meaning in Cardano discource)

Currently `approvedBidders` is just list fixed at auction announcement (see #208).

Idea is to allow use auction by issuing approval dynamical instead. Approval should be in form of signed pair `(bidder, auctionId)`.

Off-chain transfer of approval datums is out of scope of this task.

Tasks:

- [ ] Add approval datum check on-chain
- [ ] Add approval datum creation function
- [ ] Add Frontend CLI support (command for issuing and argument for using, I guess)

nazrhom:
Fixes: #220 
Fixes: #203
Fixes: #221


src-utils/HydraAuctionUtils/Types/Natural.hs
```diff
@@ -58,6 +60,10 @@ intToNatural x
 naturalToInt :: Natural -> Integer
 naturalToInt (Natural i) = i
 
+{-# INLINEABLE toBuiltinBytestring #-}
+toBuiltinBytestring :: Natural -> BuiltinByteString
+toBuiltinBytestring = serialiseData . toBuiltinData . naturalToInt
```
uhbif19, line 65: :fire: 

cc @GeorgeFlerovsky

doc/on_chain_spec.md
```diff
@@ -669,3 +669,181 @@ that distributing their respective portions of the auction fee
...
+```haskell
+validNewBid
+  :: AuctionTerms
+  -> CurrencySymbol
+  -> StandingBidState
+  -> StandingBidState
+  -> Bool
+validNewBid AuctionTerms{..} voucherCS oldStandingBidState newStandingBidState =
+  isJust $ do
+    let StandingBidState oldApprovedBidders oldBid = oldStandingBidState
```
uhbif19:
```suggestion
    let StandingBidState oldBid = oldStandingBidState
```

doc/on_chain_spec.md
```diff
@@ -669,3 +669,181 @@ that distributing their respective portions of the auction fee
...
+    let StandingBidState newApprovedBidders newBid = newStandingBidState
+    BidTerms bidderPKH bidderVK newPrice bidderSignature sellerSignature <- newBid
+    -- The seller has allowed the bidder to participate in the auction
+    guard $ verifyEd25519Signature (sellerVK terms) (sellerSignatureMessage voucherCS bidderPKH bidderVK) sellerSignature
+    -- The bidder has correctly signed the datum
+    guard $ verifyEd25519Signature bidderVK (bidderSignatureMessage voucherCS newPrice bidderPKH) bidderSignature
+    -- The new bid respects the starting price and minimum bid increment
+    -- conditions in the auction terms.
+    guard $ case oldBid of
```
uhbif19: Please separate BidTerms subfunction, cuz it will probably be used offchain.

Also it should return error reason datatype of something likethat - both for offchain and for on-chain errors.

uhbif19: I plan to do same with `validAuctionTerms` for off-chain usage.

src/HydraAuction/OnChain/Deposit.hs
```diff
@@ -43,7 +43,7 @@ mkDepositValidator (StandingBidAddress standingBidAddr, EscrowAddress escrowAddr
       -- Standing bid winner is not current bidder
       && traceIfFalse
         "Standing bid winner must not be current bidder"
-        ((bidBidder <$> standingBid (standingBidState standingBidReferenceInputDatum)) /= Just (bidDepositBidder datum))
+        ((bidderPKH <$> standingBid (standingBidState standingBidReferenceInputDatum)) /= Just (bidDepositBidder datum))
```
uhbif19, line 51: :+1:

src/HydraAuction/Delegate.hs
```diff
@@ -133,7 +134,8 @@ delegateFrontendRequestStep (clientId, request) =
         Initialized headId (Open {standingBidTerms}) -> do
           validatingAuctionTerms headId auctionTerms $ do
             let newBidTerms = standingBid $ standingBidState datum
-            if validNewBidTerms auctionTerms standingBidTerms newBidTerms
+                newVoucherCS = unVoucherCS $ standingBidVoucherCS datum
+            if validNewBidTerms auctionTerms newVoucherCS standingBidTerms newBidTerms
```
uhbif19, line 136: Why you unwrap `VoucherCS`? `validNewBidTerms` requires concrete CS only from Voucher.

src/HydraAuction/OnChain/StandingBid.hs
```diff
@@ -123,14 +121,32 @@ mkStandingBidValidator terms datum redeemer context =
...
-    Just oldBidTerms ->
-      traceIfFalse "Bid increment is not greater than minimumBidIncrement" $
-        bidAmount oldBidTerms + minimumBidIncrement terms <= bidAmount newBidTerms
-    Nothing ->
-      traceIfFalse "Bid is not greater than startingBid" $
-        startingBid terms <= bidAmount newBidTerms
-validNewBidTerms _ _ Nothing =
+{-# INLINEABLE validNewBidTerms #-}
+validNewBidTerms :: AuctionTerms -> CurrencySymbol -> Maybe BidTerms -> Maybe BidTerms -> Bool
+validNewBidTerms terms voucherCS oldBid (Just newBidTerms) =
```
uhbif19: But VoucherCS is derivable from AuctionTerms. It could be provided as param to `mkStandingBidValidator`.

src/HydraAuction/OnChain/StandingBid.hs
```diff
@@ -123,14 +121,32 @@ mkStandingBidValidator terms datum redeemer context =
...
+  -- The seller has allowed the bidder to participate in the auction
+  traceIfFalse "User is not an autorised bider" (verifyEd25519Signature (sellerVK terms) sellerMessage sellerSignature)
```
uhbif19:
```suggestion
  traceIfFalse "User is not an authorised bider" (verifyEd25519Signature (sellerVK terms) sellerMessage sellerSignature)
```

src/HydraAuction/OnChain/StandingBid.hs
```diff
@@ -123,14 +121,32 @@ mkStandingBidValidator terms datum redeemer context =
...
+    && case oldBid of
+      Just oldBidTerms ->
+        traceIfFalse "Bid increment is not greater than minimumBidIncrement" $
+          bidAmount oldBidTerms + minimumBidIncrement terms <= bidAmount newBidTerms
+      Nothing ->
+        traceIfFalse "Bid is not greater than startingBid" $
+          startingBid terms <= bidAmount newBidTerms
+  where
+    BidTerms bidderPKH bidderVK newPrice bidderSignature sellerSignature = newBidTerms
+    auctionId = voucherCS
```
uhbif19: ?

app/CLI/Actions.hs
```diff
@@ -266,13 +264,14 @@ handleCliAction sendRequestToDelegate currentDelegateStateRef userAction = do
               <> " ADA in auction "
               <> show auctionName
               <> "."
+          sellerSignature <- liftIO $ sellerSignatureForActor terms actor
```
uhbif19, line 215:
```suggestion
-- FIXME: temporaral stub until Platform server
          sellerSignature <- liftIO $ sellerSignatureForActor terms actor
```

uhbif19: :+1:

doc/on_chain_spec.md
```diff
@@ -669,3 +669,181 @@ that distributing their respective portions of the auction fee
 can be done in a single transaction.
 Later on, this can be generalized in a straightforward way
 to accommodate incremental fee distribution to a larger number of delegates.
+
+# Extensions
```
uhbif19: I still think that "Extensions" should be merged in main spec.

Spec represents current desired state of project. For tracking changes history we already placed it in repo.

uhbif19: Also `ApprovedBidders` should be eliminated from spec too.

doc/on_chain_spec.md
```diff
@@ -669,3 +669,181 @@ that distributing their respective portions of the auction fee
...
+```haskell
+validNewBid
+  :: AuctionTerms
+  -> CurrencySymbol
+  -> StandingBidState
+  -> StandingBidState
+  -> Bool
+validNewBid AuctionTerms{..} voucherCS oldStandingBidState newStandingBidState =
+  isJust $ do
+    let StandingBidState oldApprovedBidders oldBid = oldStandingBidState
```
uhbif19: Seems like version in code is more recent.

uhbif19 (changes requested):
Overall seems like solves everything! Cool. Minor request though.

uhbif19:
Please mark which issues are closed/related.

src/HydraAuction/OnChain/Escrow.hs
```diff
@@ -127,17 +127,17 @@ mkEscrowValidator (StandingBidAddress standingBidAddressLocal, FeeEscrowAddress
               Just (StandingBidDatum {standingBidState}) -> case standingBid standingBidState of
                 Nothing -> traceError "No bid in standing bid datum"
                 (Just bidTerms) ->
-                  traceIfFalse "Not signed by bidder" (txSignedBy info (bidBidder bidTerms))
```
uhbif19, line 139: @GeorgeFlerovsky  Maybe we should check sign validity here too, to prevent possible statebid forgery by L2 peers.

src/HydraAuction/OnChain/StandingBid.hs
nazrhom: I don't think we can derive it onchain

doc/on_chain_spec.md
nazrhom: I think splitting it out in different sections makes it easier to understand and follow.

These changes are not necessary to run an auction, but can be viewed as extra bits we add on top. Keeping the spec separate also allows people in the future to only pick what they need from it.

As for `ApprovedBidders`, if we merge this then we should definitely eliminate it, but otherwise we can just leave it there, as the previous spec worked with that

src/HydraAuction/OnChain/StandingBid.hs
uhbif19: Oh, right, this is not function local to `mkStandingBidValidator`. Another reason to separate sign checking code from it.

doc/on_chain_spec.md
nazrhom: I propose we do it when we need it.... we probably will at some point, but there is no reason to do it earlier. It should be easy to separate this into a function if needed

doc/on_chain_spec.md
uhbif19: > I think splitting it out in different sections makes it easier to understand and follow.

Why? You are saying same thing twice and you separate this from tx specs. All our doc is based on tx specs and requirements and that is the only case then tx specs do not say what tx does.

> These changes are not necessary to run an auction, but can be viewed as extra bits we add on top.

What you mean by "not necessary to run"? Our code does not work without it, it cannot be disabled. Without approval one cannot bid.

That is how our code works and spec should describe how our code works.

> but otherwise we can just leave it there, as the previous spec worked with that

It worked and now it does not work. This is mismatch between spec and code.

doc/on_chain_spec.md
uhbif19: > we probably will at some point

It is already used at `staging`. And I always know that it will be used, because parsing tx errors for showing UX is harder.

uhbif19: And I think it better be separated anyway. One part works with `BidTerms` and other needs all other Datum.

doc/on_chain_spec.md
```diff
@@ -303,27 +307,23 @@ under the `start bidding` redeemer:
...
 ```haskell
-initStandingBidDatum :: CurrencySymbol -> ApprovedBidders -> StandingBidDatum
-initStandingBidDatum voucherCS approvedBidders = StandingBidDatum
+initStandingBidDatum :: CurrencySymbol -> StandingBidDatum
+initStandingBidDatum voucherCS = StandingBidDatum
   { standingBid = StandingBidState
-      { approvedBidders = approvedBidders
-      , standingBid = Nothing
-      }
+      { standingBid = Nothing}
```
uhbif19, line 322: Looks like `StandingBidState` makes no sense anymore, and could be removed as well.

src/HydraAuction/Types.hs
```diff
@@ -122,46 +122,37 @@ calculateTotalFee :: AuctionTerms -> Integer
...
-  { bidBidder :: !PubKeyHash
-  -- ^ Who submitted the bid?
+  { bidderPKH :: !PubKeyHash
+  -- ^ PubKeyHash of whoever submitted the bid?
+  , bidderVK :: BuiltinByteString
+  -- ^ Verification Key of whoever submitted the bid
   , bidAmount :: !Natural
   -- ^ Which amount did the bidder set to buy the auction lot?
+  , bidderSignature :: BuiltinByteString
+  -- ^ Represents the signed payload by the bidder
```
uhbif19, line 141: Still this comment explains nothing.

doc/on_chain_spec.md
uhbif19: Looks like you already killed it in real code. So this should be sper-vs-code desync.

doc/on_chain_spec.md
```diff
@@ -79,39 +80,47 @@ is statically parametrized by the auction terms.
...
-data ApprovedBidders = ApprovedBidders
-  { bidders :: [PubKeyHash]
-  -- ^ Which bidders are approved to submit bids?
-  }
+The idea is to verify on-chain
+that the bidder has signed their new bid,
+and is meant to prevent cases where
+a delegate could submit a false bid for an unaware bidder.
+
+To verify something on-chain,
```
uhbif19, line 96: I think this should be included in limitations.md in new Plutus section.

doc/on_chain_spec.md
```diff
@@ -79,39 +80,47 @@ is statically parametrized by the auction terms.
 Similarly, all the validator scripts in the auction are made unique
 via static parametrization on the auction terms.
 
-Between the auction announcement and the bidding start time,
-the seller fixes the list of bidders
```
uhbif19, line 83: Line about approval disappeared completely. While this paragraph probably should explain that still or it is not clear why seller will provide any signed data at all.

uhbif19 (approved):

PR #319, opened 2023-05-23
Update demo instructions in README to match M4 updates

uhbif19:


nazrhom (approved):

README.md
```diff
@@ -43,11 +43,11 @@ that will start a single cardano-node,
 and 3 instances of the delegate server.
 
 Before starting the demo we will need to build the delegate server docker images.
-This is done through nix and can be done by `make build-docker`.
+This is done inside nix shell and can be done by `make build-docker`.
 This command will build the image and load it as `hydra-auction-delegate:latest`.
 
-We recommend using the `spin-up-new-devnet.sh` script
-to run the demo and start the components,
+We recommend call the `./spin-up-new-devnet.sh 1` script
```
nazrhom, line 49: I think we should change to script to do this by default. Then we can add an optional flag to disable running the delegate servers

nazrhom: But to be honest, I don't even understand why we split it up like this in the first place.
If we are using docker in our tests, then we should use that to spin up the delegate servers as well

uhbif19: Because on L2 tests we test DelegateServer as white box. That means that we emulating its execution and should not start real servers.

That way we could reliably test/avoid concurrency issues for example.

Issue #311, opened 2023-05-22
CLI UX issues evident on testnet


uhbif19:
- [x] Tx await takes long. CLI should show duration of it action - for demo comparison clearness.
- [x] `awaitTransaction` should have timeout. It should be somehow align to network.
- [x] Slots can progress with lags. That results in flaky tx submission fails due to validation interval mismatch - cuz node thinks it in older slot. This should be handled, probably by retries in such cases.
- [x] Slot should be shown in `watch` to cover lag.
- [x] `show-address` CLI action - to check on network explorer
- [x] Fix seeding action: give less ADA (or parameterize), show consistent sums in messges, show Faucet address to user

PR #324, opened 2023-05-29
M5 Actions support in exec parser and bug fixes

uhbif19:
* M5 Actions support in parser
* Delegate logic bug fixes
* Part of #311

### DIFF
--- a/delegate-app/Main.hs
+++ b/delegate-app/Main.hs
@@ -360,7 +360,7 @@ mbQueueAuctionPhases tick delegateEvents toClientsChannel = do
         Just s -> do
           -- Preventing infinite loop on stage corner
           let secsToWait = if s == 0 then 1 else s
-          threadDelay (fromInteger * 1_000_000)
+          threadDelay $ fromInteger $ secsToWait * 1_000_000
           queueCurrentStageAndWaitForNextLoop terms
 
 main :: HasCallStack => IO ()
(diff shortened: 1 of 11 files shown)


delegate-app/Main.hs
```diff
@@ -358,7 +358,9 @@ mbQueueAuctionPhases tick delegateEvents toClientsChannel = do
       case mSecsLeft of
         Nothing -> pure ()
         Just s -> do
-          threadDelay (fromInteger s * 1_000_000)
+          -- Preventing infinite loop on stage corner
+          let secsToWait = if s == 0 then 1 else s
```
nazrhom, line 363: Are we not using this? What is the infinite loop you want to prevent btw? Maybe we could expand the comment a bit to give more details

delegate-app/Main.hs
```diff
@@ -358,7 +358,9 @@ mbQueueAuctionPhases tick delegateEvents toClientsChannel = do
...
+          threadDelay (fromInteger * 1_000_000)
```
nazrhom:
```suggestion
          threadDelay $ fromInteger $ secsToWait * 1_000_000
```

nazrhom (approved):
Looks good, there is just a minor typo that is causing the build to fail, but should be fine otherwise

delegate-app/Main.hs
uhbif19: > Are we not using this? 

Using what?

> What is the infinite loop you want to prevent btw?

Awaiting 0 seconds in case of stages intervals corner.

uhbif19: -- In case of on stage intervals corner secsToWait stutter at 0.

I was probably wrong, it should not be infinite loop, it just generates giant log on awaiting 0 seconds for over 9000 times.

PR #325, opened 2023-05-29
Refactoring: more generic intervals support and proper semi-closed intervals usage

### LINKED ISSUE #268: Support for off-chain validation and intervals for `AuctionStage` and fix boundary case

nazrhom:


src/HydraAuction/OnChain/Common.hs
```diff
@@ -30,12 +30,20 @@ minAuctionFee = 2_000_000
...
-  BiddingEndedStage -> interval (biddingEnd terms) (voucherExpiry terms)
-  VoucherExpiredStage -> interval (voucherExpiry terms) (cleanup terms)
+  AnnouncedStage -> strictTo (biddingStart terms)
+  BiddingStartedStage -> rightExclusiveInterval (biddingStart terms) (biddingEnd terms)
+  BiddingEndedStage -> rightExclusiveInterval (biddingEnd terms) (voucherExpiry terms)
+  VoucherExpiredStage -> rightExclusiveInterval (voucherExpiry terms) (cleanup terms)
   CleanupStage -> from (cleanup terms)
 
+{-# INLINEABLE strictTo #-}
+strictTo :: a -> Interval a
```
uhbif19: Move this stuff to utils, please.

src/HydraAuction/Tx/Deposit.hs
```diff
@@ -161,7 +163,7 @@ losingBidderClaimDeposit terms = do
         , outs = []
         , toMint = TxMintValueNone
         , changeAddress = bidderAddress
-        , validityBound = (Just $ biddingEnd terms, Nothing)
+        , validityBound = stageToInterval terms BiddingEndedStage
```
uhbif19: This is incorrect.

Left logic has no upper bound, while stage has.

src/HydraAuction/Tx/Deposit.hs
```diff
@@ -192,7 +194,7 @@ sellerClaimDepositFor terms bidderPkh = do
         , outs = []
         , toMint = TxMintValueNone
         , changeAddress = sellerAddress
-        , validityBound = (Just $ voucherExpiry terms, Nothing)
+        , validityBound = stageToInterval terms VoucherExpiredStage
```
uhbif19: Same.

src/HydraAuction/Tx/Escrow.hs
```diff
@@ -364,5 +366,5 @@ sellerReclaims terms = do
             ]
         , toMint = TxMintValueNone
         , changeAddress = sellerAddress
-        , validityBound = (Just $ voucherExpiry terms, Nothing)
+        , validityBound = stageToInterval terms VoucherExpiredStage
```
uhbif19: Same.

src/HydraAuction/Tx/Deposit.hs
uhbif19: Solution I had in mind was intervals over stage, maybe I formulate that in issue description badly.

While seems like stage-to-posinf is the only case we do use.

src-utils/HydraAuctionUtils/Tx/AutoCreateTx.hs
```diff
@@ -91,7 +91,7 @@ data AutoCreateParams = AutoCreateParams
   , outs :: [TxOut CtxTx]
   , toMint :: TxMintValue BuildTx
   , changeAddress :: Address ShelleyAddr
-  , validityBound :: (Maybe POSIXTime, Maybe POSIXTime)
+  , validityBound :: Interval POSIXTime
```
uhbif19, line 94: Interesting solution.

On one side we reuse polymorphic code, on the other side Interval reuse intuitively feels problems prone to me.

uhbif19 (changes requested):

src-utils/HydraAuctionUtils/Tx/AutoCreateTx.hs
uhbif19: Like quote from PR comment:

> and for v3 we scrap this Interval coming up with a better (non-backwards compatible) Interval definition.

Overall their backwards comparability policy does not inspire confidence, so I would not depend on Plutus more then necessary. But I guess it that case it should be fine.

uhbif19 (changes requested):

src/HydraAuction/Tx/Deposit.hs
nazrhom: You are right. Maybe we can just add stage-to-posinf in utils as well, and use that?

uhbif19: > we can just add stage-to-posinf in utils as well

But that is domain (stage) specific thing, no?

src/HydraAuction/OnChain/Common.hs
nazrhom: Where, we don't really have onchain module in utils. I thought common onchain code should live in `onchain/common`

src-utils/HydraAuctionUtils/Interval.hs
```diff
@@ -0,0 +1,25 @@
+{-# OPTIONS_GHC -fno-specialise #-}
```
uhbif19: Sorry for nitpicking, but we have "Plutus" module in utils, so this should be its submodule as well.

src-utils/HydraAuctionUtils/Interval.hs
```diff
@@ -0,0 +1,25 @@
...
+{-# INLINEABLE strictTo #-}
+strictTo :: a -> Interval a
+strictTo s = Interval (LowerBound NegInf True) (strictUpperBound s)
+
+{-# INLINEABLE rightExclusiveInterval #-}
+rightExclusiveInterval :: a -> a -> Interval a
+rightExclusiveInterval s s' = Interval (lowerBound s) (strictUpperBound s')
+
+{-# INLINEABLE extendIntervalRight #-}
+extendIntervalRight :: Interval a -> Interval a
```
uhbif19: Not very understanible name, while I am not sure how to name it better.

uhbif19 (approved):

PR #326, opened 2023-05-30
Basic platform server implementation and common WS server code separation refactoring

### LINKED ISSUE #295: Basic implementation for Platform server without validity checking and durable storage

uhbif19:
Closes #295

### DIFF
--- a/src/HydraAuction/Platform/Interface.hs
+++ b/src/HydraAuction/Platform/Interface.hs
@@ -257,6 +257,8 @@ instance ToJSON (EntityKind x) where
 data ClientInput entity
   = Query (EntityQuery entity)
   | Command ClientCommand
+  deriving stock (Eq, Generic, Show)
+  deriving anyclass (FromJSON, ToJSON)
 
 data ClientCommand
   = ReportAnnouncedAuction AuctionTerms
@@ -277,7 +279,10 @@ data EntityQuery entity = MkQuery
   { filters :: [EntityFilter entity]
   , limit :: Maybe Int
   }
+
 newtype EntityQueryResponse entity = MkResponse [entity]
+  deriving stock (Eq, Generic, Show)
+  deriving anyclass (FromJSON, ToJSON)
 
 deriving stock instance Entity entity => Eq (EntityQuery entity)
 deriving stock instance Entity entity => Generic (EntityQuery entity)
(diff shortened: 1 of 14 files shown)


src/HydraAuction/Platform/Interface.hs
```diff
@@ -0,0 +1,387 @@
...
+
+data HydraHead = MkHydraHead
+  { staticInfo :: HydraHeadInfo
+  , allDelegatesKnown :: Bool
+  , headDelegateState :: InitializedState
+  }
+  deriving stock (Generic, Show, Eq)
+  deriving anyclass (ToJSON, FromJSON)
+
+deriving stock instance Ord HeadId
```
nazrhom, line 118: Why is this derived separately?

uhbif19: This is orphan instance for Hydra. We need it to worm map.

Could separate that to utils module or something like that.

nazrhom: Oh sorry, thought it was for `HydraHead` :/

src/HydraAuction/Platform/Interface.hs
```diff
@@ -0,0 +1,387 @@
...
+    MkSome (EntityKind entity) (container entity)
+
+instance
+  (forall entity. Show entity => Show (container entity)) =>
+  Show (Some container)
+  where
+  show (MkSome _ container) =
+    "MkSome (" <> show container <> " )"
+
+parseSomeEntityKindJSON :: Aeson.Value -> Aeson.Parser (Some EntityKind)
```
nazrhom, line 319: Could this be `Aeson.Value -> Aeson.Parser EntityKind` instead? I think this would also simplify the FromJSON instance where we use this to parse later

src/HydraAuction/Platform/Interface.hs
```diff
@@ -0,0 +1,387 @@
...
+Entity class describes their queriable behaviour.
+
+EntityKind is singletone-like functionalization for Entities,
+used to uniform queries and storage access.
+
+SomeX datatypes use existencial types to wrap EntityKind usage
+and make possible to use single datatype to access different Entities,
+as is required for server API.
+
+GADT usage makes deriving not working which affect some design decisions.
```
nazrhom: Im curious: could you expand a bit on what those are?

src/HydraAuction/Platform/Interface.hs
uhbif19: This is orphan instance for Hydra. We need it to worm map.

Could separate that to utils module or something like that.

src/HydraAuction/Platform/Interface.hs
uhbif19: No, one cannot parse GADT, only print :D
That is like the whole main problem.
Parse function would return not `forall a. EntityKind a`, but `EntityKind x` for some concrete x type.
That is essentially `exists x. EntityKind x`, which in Haskell is emulated with existential datatypes,
`Some ...` in this case.

src/HydraAuction/Platform/Interface.hs
uhbif19: They just do not. All stock deriving breaks if you use existential types (even while in some cases it seems like possible).

And `Generic` deriving is required to `anyclass ToJSON` deriving. But it will break if you have any non-derivable types inside. So using non-derivable existentials on top level makes us use deriving, while is less correct as type-encoding.

nazrhom (approved):

PR #327, opened 2023-05-30
Add show actors with min deposit to CLI

nazrhom:


test/EndToEnd/CLI.hs
```diff
@@ -128,6 +128,8 @@ depositTest = mkAssertion $ do
 
   waitUntil $ biddingStart terms
 
+  handleCliActionWithMockDelegates $ ShowActorsMinDeposit auctionName (fromJust $ intToNatural 5_000_000)
```
uhbif19: I think this line does not check anything.

uhbif19 (approved):

test/EndToEnd/CLI.hs
nazrhom: Ops, I was just using that to give it a quick test, but better not have it in there

PR #338, opened 2023-06-06
Support time on L2

### LINKED ISSUE #312: Fix BlockchainParams query for Hydra and support time on L2
Currently time on L2 is not supported, because L2 does not share slotno.
I guess we should take in from L1.

BlockchainParams is not perfectly createdh now as well. From discussion with Hydra team:

* That means that we are wrong at using queryTimeHandle and shoud use mkTimeHandle instread.
* Protocol params can be different we do not handle that.
* systemStart and eraHistory are relevant for time calculations. Using the one from L1 will be the right thing to do as we just implemented handling validation of timed transactions in https://github.com/input-output-hk/hydra/issues/196, where we use the same slot as on L1.
* stakePools have no meaning on L2.. do whatever you want there
Then they should be mempty I guess.

Source: https://mlabs-corp.slack.com/archives/C02JE7W7A8L/p1683738424212649?thread_ts=1683738272.808669&cid=C02JE7W7A8L

uhbif19:
Close #312

### DIFF
--- a/src-utils/HydraAuctionUtils/Hydra/Runner.hs
+++ b/src-utils/HydraAuctionUtils/Hydra/Runner.hs
@@ -25,11 +25,7 @@ import Test.HUnit.Lang (HUnitFailure)
 
 -- Hydra imports
 import Cardano.Api (NetworkId (Testnet))
-import Hydra.Cardano.Api (
-  NetworkMagic (NetworkMagic),
-  pattern TxValidityNoLowerBound,
-  pattern TxValidityNoUpperBound,
- )
+import Hydra.Cardano.Api (NetworkMagic (NetworkMagic))
 import Hydra.Chain.Direct.State ()
 import HydraNode (
   HydraClient,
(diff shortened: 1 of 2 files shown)


src-utils/HydraAuctionUtils/Hydra/Runner.hs
```diff
@@ -130,12 +130,14 @@ instance MonadTrace HydraRunner where
...
+    params <- liftIO $ executeL1Runner l1Context queryBlockchainParams
+    protocolParameters <- liftIO readHydraNodeProtocolParams
+    return $ params {protocolParameters}
 
-  -- FIXME: Hydra started to support time but we still do not
-  convertValidityBound _ =
-    return (TxValidityNoLowerBound, TxValidityNoUpperBound)
+  convertValidityBound x = do
+    MkHydraExecutionContext {l1Context} <- ask
+    liftIO $ executeL1Runner l1Context $ convertValidityBound x
```
nazrhom, line 140: Does hydra not provide a way to query this directly? In this case we are using our connection to the cardano-node to do the time conversion, but could we use something like https://hydra.family/head-protocol/api-reference#schema-UTCTime to query the time as the hydra-node sees it instead?

nazrhom:
Nice :fire:

nazrhom (approved):

src-utils/HydraAuctionUtils/Hydra/Runner.hs
uhbif19: No, they do not provide a way, there is no way to get SlotNo from Hydra API now.

See: https://github.com/input-output-hk/hydra/pull/858#issuecomment-1563028483

Issue #331, opened 2023-06-05
Add support for Deposit tracking and subscriptions in Platform server


uhbif19:
This could be beneficial for UX

PR #340, opened 2023-06-06
Add BidderDeposit entity to Platform server

uhbif19:
Part of #331


nazrhom:
The PR looks good, but I don't recall what was the reason why we wanted to add this to the platform server. We already have code to query deposits on L1 in the CLI

uhbif19:
> The PR looks good, but I don't recall what was the reason why we wanted to add this to the platform server. We already have code to query deposits on L1 in the CLI

Because query L1 may be slow in production and because polling/subscribing to Platform server make better UX with notification on each new deposit.

nazrhom (approved):

PR #353, opened 2023-06-12
Refactor runners and monad actor

### LINKED ISSUE #248: Move `Runner`s code and separate Actor from Runner
- [x] Move all `(Hydra)Runner` stuff to utils library
- [x] Rename `Runner` (like `L1Runner` or something like that) and check that our naming for Runners is consistent.
- [x] Derive `MonadBaseControl` and use it where is possible
- [x] Introduce `MonadActor` and `WithActorT` - to separate them from Runner. Now there are multiple `Runner` functions which does not actually use `Actor`, so we should respect that in types.
- [x] Introduce `MonadComposed`
- [x] Clarify and document our Client starting logic in tests and working code

uhbif19:
Closes #248 

Main changes:

* Now `L1Runner` does not have actor. One need to add it with `WithActorT`.
* CompositeRunner removed in favour of HydraRunner. Currently any Hydra client does have Actor and L1 client in any way.


src-utils/HydraAuctionUtils/Composite/Runner.hs
```diff
@@ -1,67 +1,11 @@
...
-  liftIO $ executeL1Runner l1Context action
-
-instance MonadHydra CompositeRunner where
-  sendCommand = runHydraInComposite . sendCommand
-  waitForHydraEvent' timeout =
-    runHydraInComposite . waitForHydraEvent' timeout
 
-instance MonadHasActor CompositeRunner where
-  askActor = runL1RunnerInComposite askActor
+type CompositeRunner = HydraRunner
```
nazrhom, line 11: Maybe we can drop the name composite runner and just use hydrarunner everywhere

nazrhom (approved):

src-utils/HydraAuctionUtils/Composite/Runner.hs
uhbif19: Yes, that is the plan, this is temporary.

PR #359, opened 2023-06-13
Testnet demo reproducing scripts and docs

uhbif19:
Blocked by #353


scripts/spin-up-new-devnet.sh
```diff
@@ -2,39 +2,67 @@
...
 do
   echo -n "."
   sleep 0.1
 done
 echo '. done'
 
-echo "Setup cardano-cli env to connect to started network"
-export CARDANO_NODE_SOCKET_PATH=./devnet/node.socket
+echo "Changing node.socket owner to current user"
+sudo chown $USER:$USER "$CARDANO_NODE_SOCKET_PATH"
```
nazrhom: This causes the following error on my system.

```
Changing node.socket owner to current user
chown: invalid group: ‘nazrhom:nazrhom’
```

nazrhom:
```suggestion
GROUP=$(id -gn)
sudo chown $USER:$GROUP "$CARDANO_NODE_SOCKET_PATH"
```

scripts/spin-up-new-devnet.sh
```diff
@@ -2,39 +2,67 @@
...
+
+echo 'Waiting for cardano node sync ..'
+
+SYNC_PROGRESS=""
+while ! [ "$SYNC_PROGRESS" = '"100.00"' ]
+do
+  SYNC_PROGRESS="$(cardano-cli query tip --testnet-magic $TESTNET_MAGIC | jq '.syncProgress')"
```
nazrhom: This will not work when running the demo from docker, as you can't query the node from outside a docker container (on osx).

scripts/spin-up-new-devnet.sh
uhbif19: Something mac-specific I belive.

scripts/spin-up-new-devnet.sh
uhbif19: >This will not work when running the demo from docker, as you can't query the node from outside a docker container (on osx).

You can run `cardano-cli` inside `cardano-node` I believe. While I forget to do that.

uhbif19: Will fix.

PR #372, opened 2023-06-21
Add jekyll theme to github pages and add section on path to commercia…

nazrhom:
…lization


uhbif19 (changes requested):
Please separate this PR into Jekyll and path_... part.

Could not Jekyll part done without `.rb`?

"Path to a full commercial solution" seems like thing @GeorgeFlerovsky know how to review, so I am leaving it to him.

Issue #54, opened 2023-02-08
Use codes for trace errors in on-chain scripts


uhbif19:
This is best practice to use codes for making smaller script sizes. 

This is really probably blocked until the end of project, to ease debugging of on-chain scripts before.

For that we need to:

* Change all traces to codes
* Add documentation page with translation of codes to messages

PR #392, opened 2023-11-27
Gf/add contract types

GeorgeFlerovsky:


hydra-auction-error/src/HydraAuction/Error/Types/AuctionState.hs
```diff
@@ -12,7 +12,7 @@ import HydraAuction.Error.Types.BidTerms (BidTerms'Error (..))
 data Buyer'Error
   = Buyer'Error'EmptyStandingBid
   | Buyer'Error'BuyerVkPkhMismatch
-  | Buyer'Error'InvalidBidTerms [BidTerms'Error]
+  | Buyer'Error'InvalidBidTerms ![BidTerms'Error]
```
uhbif19, line 15: We have strictness by default in cabal for this.

docs/on_chain_spec.md
```diff
@@ -467,8 +467,8 @@ bidderSignatureMessage
   -> BuiltinByteString
 bidderSignatureMessage auctionId bidPrice bidderPkh =
   toByteString auctionId <>
-  toByteString bidPrice <>
-  toByteString bidderPkh
+  toByteString bidderPkh <>
```
uhbif19, line 470: Why?

cabal.project
```diff
@@ -16,7 +16,9 @@ index-state:
   , cardano-haskell-packages 2023-09-14T16:16:59Z
 
 packages:
-  hydra-auction-plutus
+  hydra-auction-error
+  hydra-auction-offchain
+  hydra-auction-onchain
```
uhbif19, line 21: Why?

hydra-auction-onchain/src/HydraAuction/Onchain/Types/AuctionState.hs
```diff
@@ -0,0 +1,159 @@
...
+      bt'BidPrice - totalAuctionFees auTerms
+  | otherwise = 0
+
+-- -------------------------------------------------------------------------
+-- Plutus instances
+-- -------------------------------------------------------------------------
+PlutusTx.unstableMakeIsData ''AuctionEscrowState
+PlutusTx.unstableMakeIsData ''StandingBidState
+
+instance Eq AuctionEscrowState where
```
uhbif19, line 143: Why not `deriveEq`?

uhbif19 (commented):
Except for error codes fixing #54 I do not understand what is happening in those changes.

`DelegateInfo` seems like something for follow up project? I am not sure if on chain logic was changed or just refactoring.

hydra-auction-error/src/HydraAuction/Error/Types/AuctionState.hs
GeorgeFlerovsky: Yes, and apparently it causes PlutusTx plugin errors to have `!` strictness annotations in data types.
https://github.com/input-output-hk/plutus/issues/4193

docs/on_chain_spec.md
GeorgeFlerovsky: Currency symbol and pkh are more likely to be fixed size, while bid price byte length varies by user.
Less likely to be abusable, if we put the price at the end.

cabal.project
GeorgeFlerovsky: I want to impose a fairly strict separation:

- `hydra-auction-onchain` uses only Plutus and not Cardano API, with very few isolated unavoidable exceptions.
- `hydra-auction-offchain` uses only Cardano API and not Plutus, with very few isolated unavoidable exceptions.
- `hydra-auction-error` uses neither Plutus nor Cardano API. It defines error types that can be re-used by both on-chain and off-chain. Also, error types need to be in separate modules because I splice error codes into onchain code via template haskell (which requires everything to be imported from outside the module).

In this way, I never have to worry about PlutusTx problems in off-chain code and I never have to deal with Cardano API complexity in on-chain code. Also, this separation makes it easier for others to translate the codebase into Plutarch + CTL, which necessarily can't cross-import from their corresponding onchain/offchain libs.

hydra-auction-onchain/src/HydraAuction/Onchain/Types/AuctionState.hs
GeorgeFlerovsky: 1. `Eq` instances are easy to define and visually verify.
2. Eliminate dependencies.
3. The template haskell machinery for our custom `Eq` derivation has broken in the past when we updated dependency versions. I want to avoid this in future.

GeorgeFlerovsky:
> DelegateInfo seems like something for follow up project? I am not sure if on chain logic was changed or just refactoring.

@uhbif19 
Logic has changed. See new spec.

`AuctionInfo` and `DelegateInfo` are datums published in utxos at non-parametrized script addresses. This allows anyone to discover all delegate groups and all auctions without needing any off-chain info. I've decided that this is much preferable to our previous "platform server" solution, which was centralized and prone to breaking.

